PR #74, opened 2024-07-16
Remove messages array from /chat & /chat/:_id response

### LINKED ISSUE #62: /chat returns messages
/chat should not return any messages, because the pagination will not work for them. Instead messages should be accessed only via /chat/:_id/messages endpoint

MikhailDeriabin:
Fix #62 

1. Remove messages array from the ChatDto
2. Add serialization for /chat and /chat/:_id based on the ChatDto class
3. Update these changes in swagger


leolab1337:
I see that you also have added NoAuth decorator for the /chat on line 59 chat.controller. Was it intentional?

MikhailDeriabin:
> I see that you also have added NoAuth decorator for the /chat on line 59 chat.controller. Was it intentional?

@leolab1337, yes it is not a mistake. Previously I had to add authentication requirements for /chat and overlook, that I have made the /chat GET auth only, so I did add this small change in this branch

leolab1337 (approved):

Issue #88, opened 2024-08-01
Create a test.txt file


MikhailDeriabin:
Typical workflow (you can check the boxes if you wish to):

- [ ] Find this Issue in [Status board](https://github.com/orgs/Alt-Org/projects/1/views/2) or [My tasks](https://github.com/orgs/Alt-Org/projects/1/views/6) view
- [ ] Open this issue by clicking on its name
- [ ] Read and understand a task from the issue
- [ ] Notice that the issue has a title, description, label (type of task), assignee (who is resolving the issue) and issue number (on top with a #-prefix, unique)
- [ ] Drag-and-drop the Issue to the "In progress" column in [Status board](https://github.com/orgs/Alt-Org/projects/1/views/2) or change the Issue status to "In progress" in [My tasks](https://github.com/orgs/Alt-Org/projects/1/views/6) list
- [ ] Create a new branch from the dev branch **(NOT main)** in your IDE or [here on GitHub](https://github.com/Alt-Org/Altzone-Server/branches). The branch name must have a unified form containing ([more about branches](https://github.com/Alt-Org/Altzone-Server/wiki/dev-Branches#branches-naming)):
 - _label_  same as the issue label, tells what kind of task the branch is resolving
 - _name_ 2-5 words name of choice, which should describe the task to be resolve. Name is used for identifying the branches between each other
 - _issue number_ same as the issue number
 - Example: feature/ProfileService-read-method-45 (adds new read method to ProfileService class)
- [ ] Checkout to the created branch in your IDE
- [ ] Create a test.txt file in /src/clan folder
- [ ] Write "Hello world" to the test.txt file
- [ ] Now the required code is written and the task is done. Create a new pull request to the dev branch. You can do it [here on GitHub](https://github.com/Alt-Org/Altzone-Server/compare) as follows:
 - Choose the dev branch as _base:_ (on left hand side) and this branch as _compare:_ (on the right hand side)
 - Click the "Create pull request" button and the pull request view will be open
 - Write a title of the pull request. It should describe in general what has been done in this branch
 - Write a description in case you have some questions related to the issue or you was not able to make some of the requested tasks
 - On the right hand side you can add additional information regarding the pull request.
 - Press the _Reviewers_ and choose me MikhailDeriabin as a reviewer from the list
 - Press the _Assignees_ and choose your account from the list
 - Press the _Label_ and choose the same label as your Issue has
 - Finally press the "Create pull request" button to publish your request
[...]

PR #93, opened 2024-08-02
Create test.txt

apulkGIT:


MikhailDeriabin (approved):
Good job!

Notice however the branch naming, which consists of 3 parts:
- _branch type_ same as the issue label
- _branch name_ 2-5 words name of your choice explaining what branch does
- _issue number_ the issue number branch is resolving

So in that case the branch name could be something like that:
feature/test-file-88

You can read more about branches naming [here](https://github.com/Alt-Org/Altzone-Server/wiki/dev-Branches#branches-naming)



Closes #88

PR #178, opened 2024-09-10
Feature/item-moving-logic-112

PlayJeri:
This PR implements the requirements of the issue 112 with some changes.

1. Combined service method
    - A new method moveItems handles moving items to stock and to room
 
2. Endpoint implementation
    - Changed HTTP verb of /item/move to PATCH to more accurately describe the behaviour.

### DIFF
--- a/src/Room/room.module.ts
+++ b/src/Room/room.module.ts
@@ -1,4 +1,4 @@
-import { Module } from "@nestjs/common";
+import { forwardRef, Module } from "@nestjs/common";
 import { MongooseModule } from "@nestjs/mongoose";
 import { ModelName } from "../common/enum/modelName.enum";
 import { RoomController } from "./room.controller";
@@ -18,7 +18,7 @@ import { ItemModule } from "../item/item.module";
         MongooseModule.forFeature([ {name: ModelName.CLAN, schema: ClanSchema } ]),
         PlayerModule,
         RequestHelperModule,
-        ItemModule
+        forwardRef(() => ItemModule),
     ],
     controllers: [ RoomController ],
     providers: [ RoomService, RoomHelperService ],
--- a/src/Room/room.service.ts
+++ b/src/Room/room.service.ts
@@ -1,4 +1,4 @@
-import { Injectable } from "@nestjs/common";
+import { forwardRef, Inject, Injectable } from "@nestjs/common";
 import { InjectModel } from "@nestjs/mongoose";
 import { Model } from "mongoose";
 import { Room } from "./room.schema";
@@ -17,7 +17,7 @@ export class RoomService {
     public constructor(
         @InjectModel(Room.name) public readonly model: Model<Room>,
         private readonly roomHelper: RoomHelperService,
-        private readonly itemService: ItemService
+        @Inject(forwardRef(() => ItemService)) private readonly itemService: ItemService
     ){
         this.refsInModel = [ModelName.ITEM, ModelName.SOULHOME];
         this.basicService = new BasicService(model);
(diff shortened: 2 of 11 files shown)


MikhailDeriabin (changes requested):
Overall nice work. Here are some minor things to change:

1. swagger.json
   - It is better not to change the swagger.json file as well as any other files, 
    which are outside of the `src` folder unless it is separately written in the Issue, due to possible conflicts or any unexpected behavior.
    In case there is a need to change a swagger, it is better to mention that in the PR and I will change it.
   - You are right that the _PATCH_ method is more suitable for the request according to REST rules, however the methods for such requests are _POST_ in this API.
    The logic behind it is that this is a special endpoint, which may have only one method (unlike for ex. _/clan_) and its behavior is determined by the verb _move_.
    It is better to use one method in all API, because it is easier for the game team to use this method for such requests.

2. moveItem.dto.ts
   - _Line 13_: The field should be named `moveTo` not `move_to` (camel case). The snake case is used only for _id fields to highlight that the field is a mongo id field.

3. item.controller.ts:
   - _Line 25_ There are `@LoggedUser()` decorator to retrieve the user. Please use it instead of retrieving it straight from the request object.
   - _Line 26_ You can also not return anything from the controller method, in that case the 204 will be returned automatically as it is done in the _/characterClass_ example:
  ```ts
    @Put()
    @Authorize({action: Action.update, subject: UpdateCharacterClassDto})
    @UniformResponse()
    public async update(@Body() body: UpdateCharacterClassDto){
        //Usually you do not want to return anything for put and delete if it was success (204 response)
        const [resp, errors] = await this.service.updateOneById(body);
        if(errors)
            return [null, errors];
    }
  ```
  - When I am logged-in as a clan admin and try to move an item being in the stock to the same stock ( = not moving it anywhere), the API returns 404. It should return 204:
  ```
    //Item:
    {
        "objectType": "ItemDto",
        "_id": "66d70d08941f1d7be5033a42",
        "name": "tyyny",
        "weight": 1,
        "recycling": "common",
        "unityKey": "tyyny",
        "price": 1,
        "location": [
            1,
            2
        ],
        "isFurniture": false,
        "stock_id": "66d70d08941f1d7be5033a3f",
        "room_id": null
    }

    //Request:
    {
        "item_id": "66d70d08941f1d7be5033a42",
        "move_to": "stock",
        "destination_id": "66d70d08941f1d7be5033a3f"
    }
  ```

- When an item does not belong to a clan 404 should be returned instead of 400

4. moveTo.enum.ts:
   - The enum values must start with capital letters "Stock" and "Room"
5. item.service.ts:
   - _Lines 12-16_ The imports should be relative and start from the `.` or `..` not from `src`. It happens because you are probably using VS Code with default imports statement settings. You should change [this setting](https://stackoverflow.com/questions/60200359/tell-vs-code-to-always-use-relative-paths-for-typescript-auto-imports). Such imports can introduce random bugs so that the API will not start on some machines. Idk is it Nest problem or not.
   - _Line 174_ I would suggest to have `moveItem()` method with 2 arguments object with `itemId: string, destinationId: string, moveTo: MoveTo` fields and ``playerId`, it can help to read and understand code easier according to clean code principles.
   - The JSDocs in TS should not contain type annotations such as `* @param {string} itemId`. TS do it automatically, you should remove them.
   - _Line 201_ You should return an array with Errors not object even if it has one error
   - _Line 152_ I would suggest to have in `moveItems()` as well one argument

PlayJeri:
Changes have been made as requested. @MikhailDeriabin 

 While making the changes I noticed a bug.  If moveItemDto moveTo was "Room" user could assign any mongo id not just room_id to the room by using any valid mongoID in destination_id. To fix this I created getItemClanId method and changed the moveItem method a bit. 
 
 I also added item.errors.ts file to for the custom not found ServiceError.

MikhailDeriabin (approved):
You have done a good work, keep it up 🚀

This is a good idea to add less generic messages and more specific ones.

I have just changed imports in couple files, where they was not relative

PR #188, opened 2024-09-21
Feature/battle result informing endpoint 122

PlayJeri:
This PR should have everything required by issue 122.

The endpoint control flow goes like this.
    1. Check if the requester is in the winning team
    2.  If not early return with 403 status code
    3. Create Game if it doesn't already exist
    4. Return response with the steal token
  
I decided to go with the early return for losing team so not as many unnecessary DB calls would be made.

Since the request body doesn't include anything that can be used as unique identifier I solved the checking if game already exists by getting the latest Game from DB with request player_ids and checking if that game was created within 30 seconds. I assumed that games would always take at least 30 seconds so this number can be tweaked if needed.

The endpoint doesn't inform client about possible failure of creating the game document. It calls the service method to create it and then sends the response. Since the response doesn't depend on the created game document I though it should prioritise the response.


MikhailDeriabin (changes requested):
Overall everything is great, I think your algorithm of determining whether the game data is already saved is nice and should work, because it is expected that all clients will send the game results almost immediately so the 30 sec window should be enough. There is also no need for client to know about the game document creation so this is also fine.

You can ignore the conflicts I will resolve them, they came because I had to rename the _Room_ folder to _room_, since it was causing building errors on Mika's machine. Check that you have the _room_ folder renamed after merge in _dev_ branch.

Here are some small changes to do:

1. BattleResponseDto:
   - [ ] [Line 3]: rename `SoulHome_id` to `soulHome_id`
  
2. BattleResultDto:
   - [ ] [Line 5]: The type field should be an enum with one value "result" at the time, because some other data might come from the game in the future. This field is determining the shape of the request completely. For example, in future we might want to get collected diamonds amount during the battle and the request shape will be something like:
  
    ```ts
    {
        type: "diamonds",
        amount: 12 
    }
    ```
    With that being said I think you should create a separate DTO class for body objects with only `type` enum field required and then add a switch statement to controller which determines how to handle requests with different `type` fields as well as validate them accordingly. However there can be a problem with global pipe which adds validation in [main.ts 29](https://github.com/Alt-Org/Altzone-Server/blob/dev/src/main.ts#L29). I think it will filter out all fields that are not in this separate body DTO. You can override the global pipe as it is described [here](https://stackoverflow.com/questions/56583609/how-can-we-override-the-global-validation-pipe-in-nestjs) or just accept the `body` as _any_ and validate its `type` field straight in controller with _class-validator_ lib `validate()` function (it is already installed). You have validation in place in your service methods, so there is no need to validate whole body in the controller.

3. game.schema.ts:
   - [Line 12 and 15] Did you meant to have _string[]_ as a data type instead of _Player[]_ or why did you put _Player[]_?

4. gameData.controller.ts:
   - [ ] add use of the `switch` statement as it was described in second point.

5. gameData.service.ts:
   - [ ] Make all methods which you are not using in controller private.
   - [Line 55] Just a tip that you should not use parameter names like `body` in the service classes, because in theory service should not know about the context where it is used and the `body` is more an HTTP protocol thing of which only controller class should be aware of. 
   The DTOs (=Data Transfer Object) are also not a good thing to use in the services, but I started to use them much earlier before I knew about this concept of service and controller layers and they are everywhere now basically and it is hard to remove them at this point. It was pretty costly mistake. But yes it is more FYI rather than requirement.

PlayJeri:
Requested changes implemented.

The endpoint now accepts body type of any and then validates the "type" field of request in controller and after request is sent to correct handler by a switch statement.

changed Player[] to string[] in schema.

renamed SoulHome_id to soulHome_id

I did not yet make changes to the use of DTOs in the service but I do see your point. We could create "plain" types or interfaces that have same types as the DTO class. The performance differences should be negligible between these but they might be confusing about the intended use case.

MikhailDeriabin (approved):

PR #201, opened 2024-09-30
Replaced tag with labels array, updated DTOs and schema. Validated(required, max.5, enum)

sebastyijan-fi:
Did not make any changes to swagger, it is still referenced as tag.

### DIFF
--- a/src/clan/clan.schema.ts
+++ b/src/clan/clan.schema.ts
@@ -11,8 +11,8 @@ export class Clan {
     @Prop({ type: String, required: true, unique: true })
     name: string;
 
-    @Prop({ type: [String], enum: ClanLabel })
-    labels?: string[];
+    @Prop({ type: [String], enum: ClanLabel, required: true })
+    labels: string[];
 
     @Prop({ type: Number, default: 0 })
     gameCoins: number;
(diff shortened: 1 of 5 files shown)


MikhailDeriabin (changes requested):
Good work. 

Overall there was no need to remove the `tag` field for now, it is just deprecated. The reason is that the endpoint is already in use by both clients and if we change the interface by removing this field there can be some bugs or crushes especially in C#. On the other hand if we add a new field `labels`, we can just notify clients that they should not use the `tag` and instead use the `labels`. Once they make required changes, we can safely remove this field.

Here are small adjustments to do:

1. createClan.dto.ts:
   - [x] put `tag` field back:
   ```
    @IsString()
    tag: string;
   ```
   - [x] [Line 13] make `labels` of type `ClanLabel[]`

2. updateClan.dto.ts:
   - [x] put `tag` field back:
   ```
    @IsString()
    @IsOptional()
    tag: string;
   ```
   - [x] [Line 19] make `labels` of type `ClanLabel[]`
   - [x] [Lines 16-19] add `@IsOptional()` decorator for `labels`, because this DTO is for updates and user not necessary want to update this field each time

3. clan.dto.ts:
   - [ ] put `tag` field back:
   ```
    @Expose()
    tag: string;
   ```

4. clan.schema.ts:
   - [ ] put `tag` field back:
   ```
    @Prop({ type: String })
    tag: string;
   ```

MikhailDeriabin:
Could you also move the _ClanLabel_ enum to _/clan/enum_ folder

sebastyijan-fi:
All requested changes have been made:

- Re-adding the tag field in createClan.dto.ts, updateClan.dto.ts, clan.dto.ts, and clan.schema.ts.
- Updating the labels field to use ClanLabel[].
- Moving the ClanLabel enum to the /clan/enum folder.

MikhailDeriabin (approved):

PR #211, opened 2024-10-01
Feature/item stealing 111

PlayJeri:
Add: 
    1. /item/steal GET and POST routes as defined in the issue.
    2. Guard, decorator and type for Steal Token.
    3. Verify token method in auth service
    4. DTO for item stealing
    5. Error type for id mismatch

Change:
    1.  item service methods to private
    2. service method return types
    3. jsDocs for service methods

The commit for item.servic.ts is a bit messy looking. For some reason it shows over 200 deletions when in reality there is about 20. I had autoformat on save enabled by accident and it did something weird to the line endings maybe. There are two added methods at the end of the service stealItems and getItemSoulHomeId. Those are the only things added to the file and I changed the moveItems method so that it returns the items moved.

### DIFF
--- a/src/auth/auth.module.ts
+++ b/src/auth/auth.module.ts
@@ -10,6 +10,7 @@ import {RequestHelperModule} from "../requestHelper/requestHelper.module";
       ApiStateModule
   ],
   providers: [AuthService],
-  controllers: [AuthController]
+  controllers: [AuthController],
+  exports: [AuthService]
 })
 export class AuthModule {}
--- a/src/auth/auth.service.ts
+++ b/src/auth/auth.service.ts
@@ -1,4 +1,4 @@
-import {Injectable} from '@nestjs/common';
+import {Injectable, UnauthorizedException} from '@nestjs/common';
 import {MongooseError} from "mongoose";
 import {JwtService} from "@nestjs/jwt";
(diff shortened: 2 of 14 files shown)


MikhailDeriabin (changes requested):
It seems to be working fine 👍

I have added an exception filter for APIErrors to this branch and changed the StealToken guard to throw APIError directly.

Here are adjustments to be made:

1. /item/steal POST
   - [x] When I am logged-in with loosed user account and using _steal_token_ for won team, I do not get 403, but 400

   - [ ] When I get "no movable items found" 404 should be returned not 400

2. StealToken.decorator.ts:
   - [ ] Add JSDocs

3. StealTokenGuard.guard.ts 
   - [ ] add JSDocs

4. item.service.ts
   I would suggest to make separate service with logic for moving and stealing items. As well `ItemHelperService`, which holds helper methods like `getItemSoulHomeId()` and `getItemClanId()`. Because the `ItemService` starts to be pretty big class, which is not optimal. But it is not a requirement.

PlayJeri:
Docs added.
Incorrect status codes changed.
Item service logic moved to ItemHelper and ItemMover services

MikhailDeriabin (approved):
Nicely done!

PR #220, opened 2024-10-04
Feature/retrieve player tasks 203

PlayJeri:
The json file now has separate arrays for days from Monday to Sunday, weekly and monthly.
I wasn't sure if the Saturday and Sunday were supposed to be combined into weekend and should there already be the different options for weeks and months like week1, week2.
The PlayerTasks type is configured for the current layout of the json file but those can be easily altered later on if needed.

I also added enum for the task types. I figured that will be useful later.

The PlayerTasksService holds the tasks and loads them from the json file on module init.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,3 +113,4 @@ testAPI/**/*spec.js.map
 
 #IDEs
 .idea/
+.vscode/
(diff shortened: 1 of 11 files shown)


MikhailDeriabin (changes requested):
Looks good, please add JSDocs in:

- [x] playerTasks.service.ts, `onModuleInit()` in general what it does
- [x] task.type.ts

PlayJeri:
JSDocs added. I also moved the PlayerTasks type to separate file

MikhailDeriabin (approved):

PR #238, opened 2024-10-07
add new required fields to Clan model

MikhailDeriabin:
1. Add ageRange, goals, phrase, language fields to Clan DTOs and schema
2. Update swagger accordingly
3. Remove `gameCoins` field from Clan creation and updating DTOs, since player should not be able to update them


leolabdev (approved):
nice update , it used to be weird to hardcode this value on the frontend

leolabdev:
@MikhailDeriabin Will this update be deployed soon?

MikhailDeriabin:
@leolabdev 

I am not sure, but I think it is already in dev build (dev.altzone.fi)

MikhailDeriabin:
@leolabdev 

Now the newest version is on server devapi.altzone.fi. Although, some old clans can be without those fields and points and coins are all zeroes, because some changes need to be made in the game first.

PR #246, opened 2024-10-09
Change/add statistics to player model 231

PlayJeri:
I created gameStatistics class that extends mongoose document and embedded that into the Player schema.
I added two methods to player service to increment. played and won battles.
handleResultType in gameData service now calls those methods to track the number of battles


MikhailDeriabin (approved):
Good idea to add separate methods for handling `playedBattles` and `wonBattles` fields tracking, it should be easier to make changes to the tracing logic, if we ever need it.

I have added JSDocs for the `GameStatistics` class

PR #254, opened 2024-10-14
Feature/add played battles task 204

PlayJeri:
Core logic and schema for Tasks implemented.

I made the changes you suggested in the issue comments.
(change to error type, reverted the changes in NotificationResource enum)

The updateTaskProgress method wasn't able to set completedAt field to undefined so I use task.save() in the registerAtomicTaskCompleted instead.

Played battles get now updated from the gameData service.

### DIFF
--- a/src/common/enum/modelName.enum.ts
+++ b/src/common/enum/modelName.enum.ts
@@ -32,4 +32,5 @@ export enum ModelName {
     CLANVOTE = 'ClanVote',
     ITEMSHOP = 'ItemShop',
     GAME = 'Game',
+    PLAYER_TASK = 'PlayerTask',
 }
--- a/src/gameData/gameData.module.ts
+++ b/src/gameData/gameData.module.ts
@@ -6,7 +6,7 @@ import { Game, GameSchema } from './game.schema';
 import { PlayerModule } from '../player/player.module';
 import { ClanModule } from '../clan/clan.module';
 import { RoomModule } from '../room/room.module';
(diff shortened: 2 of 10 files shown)


MikhailDeriabin (changes requested):
Overall the core logic should work now, good job.

There should be some changes made to the schema you defined. There are some data duplication for fields `points`, `coins`, `type` and `frequency`. 
The player tasks JSON file already has this data and you do not need to save this info to DB. Compare the two variants of collection below.

image

Instead you should use the `id` field from the JSON file to determine what task it is, 
since it is guaranteed to be unique in the whole file by the person who writes tasks to the file.

I have committed couple methods to the service to the branch, which you can use to work with the predefined tasks object.

- [x] Remove duplicating fields `points`, `coins`, `type` and `frequency` from the schema, DTOs etc. and add `taskId` instead.
- [x] Change playerTasks JSON to have unique `id` fields also for week days, I guess ChatGPT can save u some time with that
- [x] Think which methods of the player task service can be private I think the `checkIfTaskIsActive()` and `getNewTaskObject()` at least
- [x] Add name definition to the schema using the `ModelName` enum as it is done for example in clan schema:
      ```ts
      TaskProgress.set('collection', ModelName.PLAYER_TASK);
      ```

PlayJeri:
I made the changes you suggested

MikhailDeriabin (approved):
Nice work 🚀

PR #256, opened 2024-10-16
Add call to taskService to track won battles

PlayJeri:
Added a call to task service with win_battle type after validating that user is in the winning team.


MikhailDeriabin (approved):
Good job.

I have noticed a bug, when some tasks are not registering, for example when player play a game and win it, where should be registered 6 tasks ( 3 played + 3 won ), but instead there was 4 tasks created in PlayerTask collection. Turned out that the mechanism of searching the existing task in `registerAtomicTaskCompleted()` was based only on frequency, but not on task type, which caused the bug.

PR #257, opened 2024-10-17
Feature/player tasks get endpoint 219

PlayJeri:
I implemented the logic for GET /playerTasks endpoint.

I created a decorator that gets the period (today, week, month) from the request and defaults to daily if not provided.
The method for getting the tasks adds the tasks (daily, weekly, monthly) based on the period value from the request.
After getting the tasks the method finds the tasks of the player who made the request from the db and updates the amount field on the task based on the task ids and amountLeft field if a task is found in the db and then returns the array of tasks.

I also noticed that the PlayerTasks type was missing Thursday so I added it to the type.

### DIFF
--- a/src/gameData/gameData.service.ts
+++ b/src/gameData/gameData.service.ts
@@ -216,7 +216,8 @@ export class GameDataService {
 		this.taskService.updateTask(user.player_id, TaskName.PLAY_BATTLE);
 		if (!playerInWinningTeam)
 			return new APIError({ reason: APIErrorReason.NOT_AUTHORIZED, message: "Invalid type field" });
-		
+
+		this.taskService.updateTask(user.player_id, TaskName.WIN_BATTLE);
 		this.playerService.addWonPlayedGame(user.player_id);
 		const [teamIds, teamIdsErrors] = await this.getClanIdForTeams([battleResult.team1[0], battleResult.team2[0]]);
 		if (teamIdsErrors)
--- a/src/gameData/gameData.service.ts
+++ b/src/gameData/gameData.service.ts
@@ -215,7 +215,7 @@ export class GameDataService {
 		this.playerService.addPlayedGame(user.player_id);
 		this.taskService.updateTask(user.player_id, TaskName.PLAY_BATTLE);
 		if (!playerInWinningTeam)
-			return new APIError({ reason: APIErrorReason.NOT_AUTHORIZED, message: "Invalid type field" });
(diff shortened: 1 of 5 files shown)


MikhailDeriabin (changes requested):
Good work, I did like your parameter decorator for retrieving requested period, it makes code cleaner.

Here are some fix that could be done:

1. period.decorator.ts
- [x] `createParamDecorator()` the switch statement is already returning the `TaskFrequency.DAILY` as a default, 
    you should remove the separate `"today"` condition from it, as well as `??` from line 12.

2. playerTasks.service.ts suggestions for refactoring:
- [ ] Line [ 70-88 ] `getPlayerTasks()` method. You can avoid duplicating `forEach` statements 
    by introducing a private method which does its logic. You can also create a method for updating amounts, 
    which gets `TaskFrequency[]` and `TaskProgress[]` as a params and returns the updated `TaskProgress[]`, 
    so that the client code could just determine which TaskFrequencies it needs to update.
- [x] `getPlayerTasks()` method could make early returns based on the TaskFrequency provided. 
    Such code could be easier to read, because currently if someone needs to check its logic 
    there is a need to read the whole method code. Here is a dummy example of this:
    ```ts
        const [tasks, errors] = await this.basicService.readMany<TaskProgress>({ filter: { playerId } });

        if (taskFrequency === TaskFrequency.DAILY){
            const updatedTasks = updateTasksAmounts(tasks, [TaskFrequency.DAILY]); //If you have implemented this method from the first point
            return [updatedTasks, null];
        }

        if (taskFrequency === TaskFrequency.WEEKLY){
            const updatedTasks = updateTasksAmounts(tasks, [TaskFrequency.DAILY, TaskFrequency.WEEKLY]);
            return [updatedTasks, null];
        }

        if (taskFrequency === TaskFrequency.MONTHLY){
            const updatedTasks = updateTasksAmounts(tasks, [TaskFrequency.DAILY, TaskFrequency.WEEKLY, TaskFrequency.MONTHLY]);
            return [updatedTasks, null];
        }
    ```

PlayJeri:
I removed the unnecessary case and period default value from the decorator.

I extracted the task amount updating logic to a separate method.

I didn't go with the way you suggested of early returns and using taskFrequency as argument. The updateAmount method takes the tasks from json file as argument and loops over all of them. But since the week or month doesn't get set unless the TaskFrequency is set to week or month the the early returns and taskFrequency checks are not necessary.

I changed the duplicate forEach loops to use nested loop instead.

MikhailDeriabin (approved):
Good job!

PR #258, opened 2024-10-18
Feature/leaderboard endpoint for top clans and players 207

PlayJeri:
Added packages required for the redis caching and setup the redis connection in the app.module.ts.

Created a new leaderboard module.

Added two endpoints /leaderboard/player and /leaderboard/clan to the controller. The endpoints support pagination and get the data from the cache if the data exists else from db.

Leaderboard service handles the caching, fetching and pagination logic.
In order to make the more generic getLeaderboard method I needed to add a method to player.service.

Added getAll method to player service to replace the deprecated readAll from the DummyServiceAbstract.
I needed to do this so that the clan and player service have uniform methods for data fetching.
I also added the publicReferences to the player.schema to make this method work.

I added points to player and clan schemas and DTO's.

The new method getAll in the player service is not strictly necessary but it makes the logic in the leaderboard service simpler. I don't know if there is a plan to make all the services uniform and remove those extends BasicServiceDummyAbstracts but the new method is created in a same way that it is in the clan service so it should be uniform with that new way if the changes are made at some point.


MikhailDeriabin (changes requested):
Overall good job as always. 

There are some classes with deprecated logic, which has to be refactor at some time, 
so you did right by adding the newer method implementation to _player.service.ts_.

Can you please explain why did you use the `bind()` for example on line 35 in leaderboard.service.ts. Was there some issue?
```ts
async getClanLeaderboard(reqQuery: IGetAllQuery) {
    return this.getLeaderboard(
        "clanLeaderboard",
        this.clanService.readAll.bind(this.clanService),
        reqQuery
    );
}
```

Things to change:

1. envVars.ts:
 -[ ] Please remove the `??` for redis variables. The idea behind this object is that 
    it is possible to check that all the env vars are defined in the _.env_ file and 
    throw an error if not on the API start. With this we can be sure that everyone has 
    an up-to-date _.env_ file, even if someone has changed it.

2. player.schema.ts:
   - [ ] [Line 70] You probably meant to set `ModelName.CUSTOM_CHARACTER` not `ModelName.CHARACTER_CLASS`. The last one is not really in use.

3. leaderboard.service.ts:
   - [ ] [Line 80] Please throw only ServiceErrors from the service classes 
    for the interface consistency, here it can be with reason `UNEXPECTED`
   - [ ] [Line 89-92] Also a `ServiceError` should be thrown, there are also the `NOT_FOUND` reason available

PlayJeri:
The binds are necessary to include the correct context when the methods are called in the getLeaderboard method. I am not aware of a better way to do this but if you have a better solution I'd like to know about it.

I changed the wrong ModelName in the player schema

I changed the thrown error types to be service errors

MikhailDeriabin (approved):

PR #260, opened 2024-10-19
Feature/add won battles reward 209

PlayJeri:
I combined the addPlayedBattles and addWonBattles to a handlePlayedBattle method.
The method now checks if the player won the battle and increments the points, played and won battles.
This simplifies the necessary method calls in the gameDataService.

This method will work when the leaderboard PR is merged since the points updating relies on the points field of the player schema that is added in that PR.

This PR also closes the issue 210 handle played battle reward for player.


MikhailDeriabin (approved):
Everything looks good

PR #262, opened 2024-10-20
Feature/add sent chat messages task 206

PlayJeri:
Created a handler method for the createMessage endpoint that calls the createMessage method and if the message is created successfully the update on player tasks service is called to update the write chat message tasks.

Added LoggedUser decorator to the endpoint to get the playerId for the task updating.

Changed the incorrect explicit return type of the createMessage method


MikhailDeriabin (approved):
Looks good

PR #298, opened 2024-10-25
Feature/add sent messages reward 208

PlayJeri:
Created a message schema to track the dates and amount of player messages and extended the gameStatistics with it.

Created a method to player service that gets the player with id and finds todays messages and increments the count by one.
If the message is not found a new one is created and if the amount is 3 player is granted 20 points.

Added a call to this new function from chat services create message handler.

### DIFF
--- a/src/authorization/rule/playerRules.ts
+++ b/src/authorization/rule/playerRules.ts
@@ -19,7 +19,11 @@ export const playerRules: RulesSetterAsync<Ability, Subjects> = async (user, sub
     if(action === Action.create || action === Action.read){
         can(Action.create_request, subject);
 
-        const publicFields = ['_id', 'name', 'uniqueIdentifier', 'profile_id', 'clan_id', 'Clan', 'CustomCharacter', 'above13', 'parentalAuth'];
+        const publicFields = [
+            '_id', 'name', 'uniqueIdentifier', 'profile_id', 'clan_id', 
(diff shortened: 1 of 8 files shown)


MikhailDeriabin (changes requested):
Overall it seems to be working fine, but I have found one bug.

Also I fixed the issue with the `gameStatistics` and `points` not being showed in /player GET. I have committed the changes to this branch.

Things to do:

- [x] I have found a bug while trying to send a message:
```
C:\Users\mikhail\Desktop\folders\private_OMEN\projects\Altzone-Server\src\player\player.service.ts:179
            player.gameStatistics.messages.push(newMessage);
                                           ^
TypeError: Cannot read properties of undefined (reading 'push')
    at PlayerService.trackPlayerMessageCount (C:\Users\mikhail\Desktop\folders\private_OMEN\projects\Altzone-Server\src\player\player.service.ts:179:44)    
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```
The error is thrown when in player's `gameStatistics` is only `playedBattles` and `wonBattles` fields. I guess mongoose does not change this object automatically, at least for existing players.

PlayJeri:
Found the bug. I had the new message being pushed to the player.gameStatistics instead of the messages array. I Tested it by removing gameStatistics from player completely and by deleting the messages from players gameStatistics.

MikhailDeriabin (approved):
Good work

PR #306, opened 2024-10-27
add LoggerUser and player_id filter to getAll

PlayJeri:
I added LoggerUser decorator to the getAll endpoint and use the player_id from that as a filter in the readAll query.


MikhailDeriabin:
@PlayJeri 

tasks of CustomCharacter milestone were not meant for you, but for Mika, as a training tasks. 

Please ask me of the tasks that you want to do first, if you have not any

MikhailDeriabin:
I guess it is better to let these tasks to be undone, because I can not really think of any other simple tasks there are. Unless you know some? They must be easy and step-by-step defined, kinda "nothing can go wrong" type.

PlayJeri:
We can leave this undone. I just saw that these issues were not assigned to anyone. I will ask you in the future

PR #347, opened 2024-11-26
Feature/add flea market sell endpoint 197

PlayJeri:
I have added / changed the following:

1. Added sell endpoint to flea market controller.
2. Added methods to flea market service to handle the item selling logic.
3. Added voting module.
4. Added voting service with logic to start new votings.
5. Added voting and vote schemas.
6. Added voting notifier.
7. Added DTOs, enums and types.

Like we discussed in the issue comments the voting schema now has entity_id prop that is not in the ERD so that needs to be updated when this PR is merged. The entity_id is ObjectId type but since the entity can be anything I did not add a ref field to it. So now we would need to check the type of the voting to know what type of entity id it is. If you think this is not a good solution I could add a another prop. entityType with type of ModelName and specify there the entity type and with that I think it would be possible to create the virtual ref as well.
I also removed the validations from the schemas and the validation will be handled in the service layer.

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -27,6 +27,7 @@ import { GameEventsBrokerModule } from './gameEventsBroker/gameEventsBroker.modu
 import { RewarderModule } from './rewarder/rewarder.module';
 import { StatisticsKeeperModule } from './statisticsKeeper/statisticsKeeper.module';
 import { FleaMarketModule } from './fleaMarket/fleaMarket.module';
+import { VotingModule } from './voting/voting.module';
 
 // Set up database connection
 const mongoUser = envVars.MONGO_USERNAME;
@@ -66,6 +67,7 @@ const redisPort = parseInt(envVars.REDIS_PORT);
 
       ItemMoverModule,
       FleaMarketModule,
+      VotingModule,
 
       CustomCharacterModule,
       ClanInventoryModule,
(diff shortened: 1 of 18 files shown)


MikhailDeriabin (changes requested):
Everything seems to be working, good job

There was a bug with voting.schema definition, it was throwing a `TypeError` for votes array, 
to fix that I have changed the type from [ Vote ] to Array< Vote >.

I have also updated the ERD.

1. voting.service.ts:
  - [x] [Line 43] `startItemVoting()` should accept one object as a parameter. 
  There are too many params with same data type and it can be hard to read.

2. fleaMarket.service.ts:
  - [x] [Lines 164-185] `moveItemToFleaMarket()` method has a bit odd structure. 
  Consider not to throw errors inside `try-catch`, but instead create a method for cancelling transaction:
  ```ts
  private async cancelTransaction(session: ClientSession) {
    await session.abortTransaction();
    await session.endSession();
  }
  ```

PlayJeri:
I made the requested changes.

The startItemVoting now takes a startItemVoting type as argument and destructures the type to variables.

Refactored the moveItemToFleaMarket so that it no longer has the try-catch-finally block and instead it uses cancelTransaction helper method in case of an db error.

MikhailDeriabin (approved):

PR #351, opened 2024-12-13
Feature/add voting/:_id endpoint 199

PlayJeri:
In this PR I added a new controller for voting and endpoint for getting a voting with id and a bug fix.

I noticed that there are two voting.service.ts files and the other one was in schemas folder. 
I deleted the one in the schemas folder and added the methods from it to the one in the correct folder and then changed the import path to that in the flea market service. I also changed the itemId in startItemVoting call to be the newly created flea market item id from the old deleted item id.


MikhailDeriabin (approved):
Good job!

PR #352, opened 2024-12-15
Feature/add /voting endpoint 200

PlayJeri:
This PR includes the following changes

1. Added /voting PUT endpoint and methods to validate and the new vote to the voting document.

2. Bugfix for the voting notifier. The old payload did not match the one defined in the drawio diagram. So I changed the parameters passed by some methods and added the virtual field for flea market item to the voting schema so the notifier receives all the information it needs. I also added the payload builder helper method to the notifier.


MikhailDeriabin (approved):
Nicely done

PR #354, opened 2024-12-17
Feature/add endpoint to get all clan votings 344

PlayJeri:
In this PR I added the controller endpoint and service method to get all the votings related to the requester.

The issue described to get all the poll objects that the clan has but since the votings document has clan_id and player_id in the organizer prop I made the query filter so that it returns if either of those ids match.

Would you like me to add the swagger documentation for this endpoint or do you want to it yourself?


MikhailDeriabin (approved):
@PlayJeri 

I can update the swagger at some point

PR #355, opened 2024-12-21
Feature/add profile info endpoint 289

PlayJeri:
I added the GET /profile/info endpoint to get the same information (Profile, Player, Clan) that auth/signIn returns and a service method to get the data from db.


MikhailDeriabin (approved):
Good job

PR #359, opened 2025-01-04
Feature/add get clan leaderboard position 336

PlayJeri:
I have added / changed the following:

I added 
 - GET /leaderboard/clan/position endpoint to get the position of the requesters clan on the leaderboard.
 - Service method to get the position data
 - Indexes on Clan and Player schemas points field to make the leaderboard queries faster


MikhailDeriabin (approved):
Good work!

PR #360, opened 2025-01-05
Feature/add get clan items endpoint 345

PlayJeri:
I have added the following:

GET /clan/:_id/items endpoint.
The endpoint gets all items for a clan with provided id.
Returns not authorised error if user is not a member of the clan.

### DIFF
--- a/src/clan/clan.controller.ts
+++ b/src/clan/clan.controller.ts
@@ -35,12 +35,17 @@ import { UniformResponse } from "../common/decorator/response/UniformResponse";
 import { IncludeQuery } from "../common/decorator/param/IncludeQuery.decorator";
 import { publicReferences } from "./clan.schema";
 import { Serialize } from "../common/interceptor/response/Serialize";
+import { RoomService } from "../clanInventory/room/room.service";
+import { ItemService } from "../clanInventory/item/item.service";
+import { ItemDto } from "../clanInventory/item/dto/item.dto";
 
 @Controller('clan')
 export class ClanController {
     public constructor(
         private readonly service: ClanService,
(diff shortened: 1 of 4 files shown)


MikhailDeriabin (changes requested):
Everything is working, but could you please make two adjustments:

1. Remove the `_id` parameter from the path, since you do have information about who is logged-in, so that the path would be `/clan/items`.
   I guess you are asking for the clan `_id` to reduce DB requests, but I think later we could add the `clan_id` to the auth token and get the info of that clan.

2. Place items into different arrays `Stock` and `SoulHome`. You can do it by filtering the items, received from DB, based on the `stock_id` and `room_id` fields are null or not:

```ts
//Pseudo code
const clanItems = await service.getClanItems();
const stockItems = clanItems.filter(item => item.stock_id);
const soulHomeItems = clanItems.filter(item => item.room_id);
```

PlayJeri:
I made the changes.

The clanId is now fetched from db with the logged in user player_id.
Response type is now separated into stock and soulhome items arrays.

The UniformResponse metadata\s dataCount was not working correctly with nested arrays or multiple object arrays. Specifically, when the dataType was Object, the dataCount defaulted to 1, which did not accurately reflect the number of items in nested arrays such as stockItems and soulHomeItems.

The formatResponse function has been updated to handle nested arrays dynamically. It now iterates over the keys of the data object and sums the lengths of any arrays found. This ensures that the dataCount accurately reflects the total number of items, regardless of the specific structure of the data object.

MikhailDeriabin (approved):
ok, nice work

PR #368, opened 2025-01-14
Bug/fix tests for clan items 365

PlayJeri:
Changed the expected items in tests to match the new default items


MikhailDeriabin (approved):
Nice work

PR #370, opened 2025-01-21
Test/write tests for chat module 324

PlayJeri:
Added tests for chat service methods and isChatExists decorator function

### DIFF
--- a/src/test/chat/chatService/createMessage.test.ts
+++ b/src/test/chat/chatService/createMessage.test.ts
@@ -1,0 +1,82 @@
+import { mongo } from "mongoose";
+import { ChatService } from "../../../chat/chat.service";
+import ChatModule from "../modules/chat.module";
+import ChatBuilderFactory from "../data/chatBuilderFactory";
+
+describe("ChatService.createMessage() test suite", () => {
+    let chatService: ChatService;
+    let chatId: string;
+	const messageBuilder = ChatBuilderFactory.getBuilder('CreateMessageDto');
+    const chatModel = ChatModule.getChatModel();
+
+    /**
+     * Before all tests, create a chat to be used for message tests.
+     * This ensures that there is a chat available in the database
+     * for the messages to be created in.
+     */
+    beforeAll(async () => {
+		const chatBuilder = ChatBuilderFactory.getBuilder('CreateChatDto');
+        const testChatName = "testChat";
+        const chatToCreate = chatBuilder.setName(testChatName).build();
+        chatService = await ChatModule.getChatService();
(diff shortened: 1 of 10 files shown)


MikhailDeriabin (changes requested):
Here are some notes and fixes to do:

_configureResponse.test.ts_:

1. Better not to test private methods. 
  Usually it is enough when you test public ones, which use private. Here is why:
  when you are writing code, you usually do not change public methods, 
  when you do so you are expecting that there could be some problems in other parts of
  the code base. If it is a private method = not part of the class interface, you are not
  expecting that these methods are in use anywhere, so you can rename, delete and change them safely, 
  as long as you check that the class interface is not impacted. In case with tests
  of private methods they fail after changes

_createMessage.test.ts_:
  1. FYI: be careful with creating shared resources for the tests, since it can lead to
    [interacting tests](https://test-smell-catalog.readthedocs.io/en/latest/Dependencies/Dependencies%20among%20tests/Interacting%20Tests.html#interacting-tests).
    Although here creation of the common chat is justified.

  - [x] Please move SUT (`ChatService`) creation to before each hook, since you need to
    create a new instance of the SUT each time to avoid the interacting tests.

  2. [Line 34] : the `id` field is a number and not the mongo id, since we get it from 
    Photon chat system, which is basically a "AUTO_INCREMENT" field, but if your solution works it is ok.

   - [x] [ Line 44 ] Make the assertions for all the fields of the message not only content.
   You can use something like:
   ```ts
   expect(chat.messages[0]).toEqual(expect.objectContaining(messageToCreate));
   ```

   - [ ] Check that for all invalid inputs nothing is created in DB via DB call not only
     with method rejection, but with actual DB calls, better by adding separate test cases for them. 
     If message is created in DB, do not fix anything in the code, just add to the test file:
     ```ts
       // TODO: Actually creates the message even if input is invalid
     ```

_readAllMessages.test.ts_:

  1. You can create message objects on top of test suite and reuse them, so that there is no
    need to create them separately in each test.

  - [ ] Make assertions for all message object not only for the `content` field.


_readOneMessageById.test.ts_:

  - [ ] Create `chatService` in `beforeEach()` hook, not in the before all.
  - [ ] Make assertions for all message object not only for the `content` field.
  1. [ Line 60 ] there are a function `getNonExisting_id()`, you can use it if you want.

  _IsChatExists.test.ts_

  - [ ] Separate test cases for true and false.

PlayJeri:
I made the changes you required

MikhailDeriabin (approved):
Nice work!

PR #398, opened 2025-02-01
Feature/add task generating mechanism 376

PlayJeri:
I decided to do a milestone branch since the individual issues would've broken the dev branch. I will merge all the milestone issues to this branch and let's merge this into the dev once the milestone is complete.

So in the milestone branch I renamed playerTasks to dailyTasks, removed most of the old logic and refactored the schema. 
I added the clanId to the task schema and made playerId optional. I also added the points and coins to the schema since they are now randomly generated.

In this PR I added the required methods to generate values for new random tasks.

I added taskConstants where the min and max values for the tasks can easily bet set.

The task type is generated based on the TaskName enum so the types of tasks the api doesn't yet support need to be commented out before we take this in use.

### DIFF
--- a/src/dailyTasks/consts/taskConstants.ts
+++ b/src/dailyTasks/consts/taskConstants.ts
@@ -1,0 +1,13 @@
+export const TASK_CONSTS = {
+	AMOUNT: {
+		MIN: 2,
+		MAX: 20,
+	},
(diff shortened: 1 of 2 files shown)


MikhailDeriabin:
ok

Issue #380, opened 2025-01-24
change task completion logic


MikhailDeriabin:
### General description

When the task is completed, where should be generated a new task for the clan.

### Things to notice

The clan may have max. 10000 points, so if the clan already has this amount no new tasks should be generated

Issue #379, opened 2025-01-24
generate and save tasks on clan creation


MikhailDeriabin:
### General description

When the clan is created there should be 20 new player tasks generated for it.

Issue #377, opened 2025-01-24
add endpoint for getting task info by _id


MikhailDeriabin:
### General description

add /playerTask/:_id GET endpoint

Issue #376, opened 2025-01-24
add tasks generating mechanism


MikhailDeriabin:
### General description

The tasks are generated by the API, and should be random. The only fixed thing is the type of task, which is already defined. The example of generated task:

Play 10 games. (where number of games is randomly generated).

MikhailDeriabin:
@PlayJeri 

1. Just in case you can create a new module yes, I think better to call it `dailyTask` to be inline with the game side.
2. Sure you can add all necessary fields in your opinion including `clan_id`.
3.  Yes the _playerTasks.json_ seems not to be needed anymore. However, some kind of object with basic info about each task type may be needed. It can hold info about `points` and `coins` per each atomic task. 
I was also thinking about the timer functionality, since its duration is most likely defined by the amount of atomic tasks (play 5 games vs 10 games). So you should define some duration of each atomic task, you can just make wild guesses at this point, because it is not yet defined.
Summarizing all, there should be one object with all the configurations for each task type, so it could be easily adjusted in the future.

PS: make sure to mention the changes to endpoints, DB etc. in PR, so I can update the docs

PR #411, opened 2025-02-04
Milestone/change player tasks to daily tasks

PlayJeri:
This pull request includes a rewrite of the task system, from player tasks to daily tasks and involves renaming, restructuring of modules, services and  schemas and updates to logic and constants.

This is still a work in progress and I have not tested everything about this yet since I have some questions about the functionality.

So the tasks are now randomly generated and therefore the task schema changed to have all the info in it. It also holds a unique playerId index to make sure player can only have one task at a time.

When clan is created 20 random tasks are created for the clan.

When a task is completed a new one is generated by resetting the some fields and assigning new random values to to others.

I'm not sure if I understood the logic of the task reservation system correctly. How it works now is that player can reserve a task that is not currently reserved and belongs to the player's clan and the player has 20minutes to complete the task. If not completed in time the task is deleted and new one created. Was it supposed to work like this or should there be infinite amount of time to complete the task but you can only reserve it for yourself for 20minutes and then someone else can  continue it.

The reservation timing works on a bullmq queue just like the votings.

The tasks can be deleted if they belong to your clan and are not reserved by any other player. This works fine if the idea is that you have to complete the task in 20 minutes but if they are meant to be shared then someone could delete the task right before it's completed. So I need some info from game designers about the details.

Should the voting system be utilised for task deletion if the task has already been started?

Should there be a rate limit on the task deletion / regeneration. Players could now simply spam the deletion until they get a really easy task. Maybe there should be some limits to this?

This is what I've currently done and what questions I have about this new system.
I tried to find specifics on google drive and discord but only found info about the old system. Is there a file somewhere where these specifics are defined and if not, who should I contact?

I wasn't sure if a PR was the best option to have this conversation but here is the progress I've made. Now let's make sure I've implemented the features as requested by game design and then I test everything to make sure this PR is ready to be merged.


MikhailDeriabin:
@PlayJeri 

> When a task is completed a new one is generated by resetting the some fields and assigning new random values to to others.

Do you mean that you are not resetting the type of the task, only the numeric values of it? If so, the task should be completely random, and you should randomly choose the type of task and the numeric values for it. It is not restricted (yet at least) that there can not be 20 same types of tasks.

--------------
> I'm not sure if I understood the logic of the task reservation system correctly. How it works now is that player can reserve a task that is not currently reserved and belongs to the player's clan and the player has 20minutes to complete the task. If not completed in time the task is deleted and new one created. Was it supposed to work like this or should there be infinite amount of time to complete the task but you can only reserve it for yourself for 20minutes and then someone else can continue it.

Well almost, if player did not complete the task in time, its progress is reset and someone else can take this task:

1. Player chooses the task "Play 3 battles" and 20 min timer starts.
2. Player plays 1 battle => _1/3 battles played_
3. 20 min is over && only _1/3 battles played_ => reset the progress back to _0/3 battles played_ + free the task that other clan members can take it.

It is worth mentioning that, it is possible that the same amount of time for different tasks is not optimal (20 min for playing 2 battles vs 10 battles). It is not defined yet at all, but I am sure that in the future some adjustment mechanism will be requested (5 min per battle => 2 battles = 10 min and 10 battles = 50 min), so make sure that the 20 min time limit is not hard coded.

--------------

> The tasks can be deleted if they belong to your clan and are not reserved by any other player. This works fine if the idea is that you have to complete the task in 20 minutes but if they are meant to be shared then someone could delete the task right before it's completed. So I need some info from game designers about the details.

Yes, I think that if task is reserved it becomes read-only for others. The idea that player can delete a task, when any of the 20 tasks do not appeal to him / her and get a new one(s).

--------------

> Should the voting system be utilised for task deletion if the task has already been started?

No, at least for now. Do not add any authorization rules for it yet, since they are not defined anyway.

--------------

> Should there be a rate limit on the task deletion / regeneration. Players could now simply spam the deletion until they get a really easy task. Maybe there should be some limits to this?

No, but I think it can come up at some point. For now just keep it simple, since again nothing is defined. First I want to see this system in use on the client side then we can add extra things to it.

--------------

> I tried to find specifics on google drive and discord but only found info about the old system. Is there a file somewhere where these specifics are defined and if not, who should I contact?

🤣 Welcome to the Altzone. In drive can be found some info, but never all of it. I did put key points to [this milestone](https://github.com/Alt-Org/Altzone-Server/milestone/23). Helena put some links [here](https://github.com/orgs/Alt-Org/projects/1/views/8?pane=issue&itemId=93819893&issue=Alt-Org%7CAltzone-Server%7C369).

Overall you can ask Helena or Niko (game UI lead) about specifics.

--------------

> I wasn't sure if a PR was the best option to have this conversation but here is the progress I've made. Now let's make sure I've implemented the features as requested by game design and then I test everything to make sure this PR is ready to be merged.

Ok, the PR is fine place for it, although I think issue, that you are doing or mostly relates to the topic, could be a better one.

Let me know if you have some other questions and when you are done just leave a comment here.

PlayJeri:
@MikhailDeriabin 

> > 
> 
> Do you mean that you are not resetting the type of the task, only the numeric values of it?

Task type is also randomly reset.


I will fix the timer logic and set the time based on the task amount not just 20minutes.

I tried to do this one issue at a time but the changes were pretty significant so I had to do couple of issues at a time to start to know what is actually working. I'll do the changes and make sure everything works.

PlayJeri:
@MikhailDeriabin 

I think I tested all the new features now and I changed the logic as requested for the task reservations.
The schema now has some extra fields to the ERD needs to be updated. 
The controller is renamed from playerTasks to dailyTasks so swagger needs to be updated and there are new endpoints.
Get "/" to get clan tasks
Get "/:_id" to get task by id
Patch"/:_id" to reserve a task
Delete "/:_id" to delete a task

Below you can find info about the individual issues.

#379 ClanService createOne calls the generateTasksForNewClan in dailyTasks service and that creates 20 tasks for the new clan. The task time is the task amount doubled. So if you have play 20 matches you have 40 minutes.
#### File: `src/dailyTasks/dailyTasks.service.ts`

```typescript
/**
 * Generates a set of tasks for a new clan.
 *
 * This method creates 20 tasks with random values and assigns them to the specified clan.
 * Each task is created by calling `createTaskRandomValues` and then adding the `clanId`.
 * The tasks are then saved using the `basicService.createMany` method.
 *
 * @param clanId - The ID of the clan for which tasks are being generated.
 * @returns A promise that resolves to the result of the `createMany` operation.
 */
async generateTasksForNewClan(clanId: string) {
  const tasks: Partial<Task>[] = [];
  for (let i = 0; i < 20; i++) {
    const partial = this.createTaskRandomValues();
    const timeLimitMinutes = partial.amount * 2;
    const task: Partial<Task> = {
      ...partial,
      amountLeft: partial.amount,
      timeLimitMinutes,
      clanId,
    };
    tasks.push(task);
  }

  return await this.basicService.createMany(tasks);
}
```

#376 
### Added `generateTaskRandomValues` Method

This method generates random values for a task, including the amount, points, coins, type, and title. It is used to create tasks with random attributes. This is utilized in other methods for task generation.

#### File: `src/dailyTasks/dailyTasks.service.ts`

```typescript
/**
 * Generates random values for a task.
 *
 * @returns An object containing random values for a task.
 */
private createTaskRandomValues(): Partial<Task> {
  const amount = Math.floor(Math.random() * (TASK_CONSTS.AMOUNT.MAX - TASK_CONSTS.AMOUNT.MIN + 1)) + TASK_CONSTS.AMOUNT.MIN;
  const points = Math.floor(Math.random() * (TASK_CONSTS.POINTS.MAX - TASK_CONSTS.POINTS.MIN + 1)) + TASK_CONSTS.POINTS.MIN;
  const coins = Math.floor(points * TASK_CONSTS.COINS.FACTOR);
  const taskType = this.getRandomTaskType();
  const titleString = this.getTaskTitle(taskType, amount);

  return {
    amount,
    points,
    coins,
    type: taskType,
    title: titleString,
  };
}
```

#374 
### Added DailyTaskQueue to handle the reservation duration.
```typescript
@Injectable()
export class DailyTaskQueue {
    constructor(
        @InjectQueue("daily-tasks") private readonly dailyTaskQueue: Queue
    ) {}

    async addDailyTask(task: DailyTaskDto) {
        if (!task.playerId || !task.startedAt) return;
        const delay = TASK_CONSTS.TIME * task.timeLimitMinutes; // convert minutes to ms
        await this.dailyTaskQueue.add("tasks", task, { delay });
    }
}

@Processor("daily-tasks")
export class DailyTaskProcessor extends WorkerHost {
    constructor(
        @Inject(forwardRef(() => DailyTasksService))
        private readonly dailyTasksService: DailyTasksService
    ) {
        super();
    }

    async process(job: Job<DailyTaskDto>): Promise<void> {
        console.log("Processing task expiration:", job.data);
        await this.dailyTasksService.handleExpiredTask(job.data);
    }
}
```

#375 
### Added reserveTask 

Reserves the task for the player if the playerId field is not set, the task belongs to player's clan and player doesn't not have task already reserved. The playerId duplicate is handled by the schema unique index and not by the method. This is in HTTP PATCH /dailyTaks/:_id endpoint
```typescript
/**
 * Reserves a daily task for a player.
 *
 * @param playerId - The ID of the player reserving the task.
 * @param taskId - The ID of the task to be reserved.
 * @param clanId - The ID of the clan to which the task belongs.
 * @throws Will throw an error if the task cannot be read or updated.
 * @throws Will throw an error if the task is already reserved by another player.
 */
async reserveTask(playerId: string, taskId: string, clanId: string) {
  const [task, error] = await this.basicService.readOne<DailyTaskDto>({
    filter: { _id: taskId, clanId },
  });
  if (error) throw error;
  if (task.playerId) throw taskReservedError;

  const startedAt = new Date();
  task.playerId = playerId;
  task.startedAt = startedAt;

  const [_, updateError] = await this.basicService.updateOneById(
    taskId,
    task
  );
  if (updateError) throw updateError;

  await this.taskQueue.addDailyTask(task);
  await this.notifier.taskReceived(playerId, task);
}

@Patch("/:_id")
@UniformResponse()
async reserveTask(@Param() param: _idDto, @LoggedUser() user: User) {
  const clanId = await this.playerService.getPlayerClanId(user.player_id);
  return await this.dailyTasksService.reserveTask(
    user.player_id,
    param._id,
    clanId
  );
}
```

#380 
#### File: `src/rewarder/clanRewarder/clanRewarder.service.ts`
points are added here and this line ensures that points don't go over 10k
```typescript
const newPoints = Math.min(this.clanMaxPoints, points + pointsToAdd);
```

#377 
#### File: `src/dailyTasks/dailyTasks.controller.ts`
Added HTTP GET /dailyTasks/:_id endpoint to get the details.

MikhailDeriabin (changes requested):
Overall seems to be working, I did not test the functionality manually, since we will be writing integration tests soon for it.

Small adjustments:

1. dailyTasks.schema.ts:

  - [x] [Line 11-14] MongoDB _id fields should always be represented as `_id`. Please make change the `clanId` to `clan_id` and `playerId` to `player_id`.
  - [x] You can also add the reference to the `Clan` collection here, as it is done for the `Player`.

2. dailyTaskDto.dto.ts:
   
  - [x] Rename the `clanId` to `clan_id` and `playerId` to `player_id`

3. dailyTasks.controller.ts:

  - [x] [Line 36] The update method should be PUT, for consistency reasons.

4. dailyTask.queue.ts:

  - [x] [Line 25] Consider removing the circular dependency in the constructor. It seems like the only use of the `DailyTasksService.handleExpiredTask()` method is within this class => mb move this method to `DailyTaskProcessor` or replace the code of the `process()` method with the code of the method? The `DailyTasksService.handleExpiredTask()` method can be removed completely then.

5. tasksConstants.ts:
  
  - [x] Add some short JSDocs, for example as it is done in _src/common/service/envHandler/envVars.ts_

6. dailyTasks.service.ts:

  - [x] Consider moving methods, which are related to tasks generation into separate class, something like `DailyTaskGenerator`, so that the DailyTasksService could handle all the DB calls and business logic can be in another. (separation of concerns principle).
  These methods are `getRandomTaskType()`, `getTaskTitle()`, `createTaskRandomValues()`, mb some other?

PlayJeri:
I made the requested changes

MikhailDeriabin (approved):
Good job!

src/dailyTasks/dailyTask.queue.ts
```diff
@@ -0,0 +1,71 @@
...
+import { InjectModel } from "@nestjs/mongoose";
+import { DailyTask } from "./dailyTasks.schema";
+import { Model } from "mongoose";
+import BasicService from "../common/service/basicService/BasicService";
+import DailyTaskNotifier from "./dailyTask.notifier";
+
+@Injectable()
+export class DailyTaskQueue {
+	constructor(
+		@InjectQueue("daily-tasks") private readonly dailyTaskQueue: Queue
```
PlayJeri, line 16: What

PR #457, opened 2025-03-02
Feature/endpoint to reset test session 422

PlayJeri:
### Brief description

Added endpoint and service method to reset a testing box. The reset endpoint can be used by the group admin only. The method finds the box of the admin and deletes the old one and then creates a new one based on the info of the old one and creates a new access token. This way the boxCreator can be used for resetting and no new logic is required.

### Change list

- enabled the BoxAuthGuard in box controller
- added the reset endpoint
- added cancelTransaction method to common module functions
- changed deleteBoxReferences to use a transaction.
- added getBoxResetData method to box service to get data from old box to the new one.

### DIFF
--- a/src/box/box.controller.ts
+++ b/src/box/box.controller.ts
@@ -3,10 +3,12 @@ import {
     Controller, Delete,
     Get,
     Param,
+    Patch,
     Post,
     Query,
     Req,
     Res,
+    UseGuards,
 } from "@nestjs/common";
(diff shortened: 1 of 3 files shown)


MikhailDeriabin (approved):
Looks good.

Couple of things to remember:

1. Tests, at least for the service level.
2. The default title of the PR should be changed to a bit more readable, because it goes to the release notes change list, but overall the PR description is pretty good.

PR #467, opened 2025-03-08
Added endpoint for deleting testing box data

PlayJeri:
### Brief description
Added endpoint for deleting the box and group admin and all associated data.
Added tests for the deleteBox service method.

### Change list

- Added /delete endpoint the the box controller
- Added deleteBox service method for box service
- Added unit tests for the deleteBox method
- Fixed the issue in tester.service.ts createPlayers method where the explicit return type did not match the actual one.

### DIFF
--- a/src/__tests__/box/BoxService/deleteBox.test.ts
+++ b/src/__tests__/box/BoxService/deleteBox.test.ts
@@ -18,6 +18,7 @@ import { Box } from "../../../box/schemas/box.schema";
 import { envVars } from "../../../common/service/envHandler/envVars";
 import { Environment } from "../../../common/service/envHandler/enum/environment.enum";
 import { SEReason } from "../../../common/service/basicService/SEReason";
+import ItemModule from "../../clanInventory/modules/item.module";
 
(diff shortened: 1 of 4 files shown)


MikhailDeriabin (changes requested):
Overall looks good, nice job.
Removing other things like votings, shop etc. would be a bit overkill at this point, because many things can change. So it is good that you remove only the most obvious things.

Couple of things to change:

deleteBox.tests.ts:

- [ ] [Line 165] Move the `expect` statements into different test cases. You almost always want to have only one thing tested per suite. For example here these things may be is the box removed, are the clans removed, are the soul homes removed etc.
  The reason why it is considered a better way is the feedback the developer can get from your tests. So that if after 2 months somebody will make changes to the deletion logic and something breaks, he can quickly see the exact reason what is wrong, and he can also check the failing test code, which will be really concise and fast to read. However if you put all of your test logic into one test case, the developer will have to read much more code.

- [ ] Add also a test for checking that all of the items were removed. Again preferably one test case for soul home items and one for stock.

PR #474, opened 2025-03-18
Fixed bug, where player can add multiple same type custom characters to own account

jkrizsan:
### Brief description
Validate if CustomCharacter is already exists for _/customCharacter_ POST endpoint. Return with 409 error code if yes. Small code quality improvements.

### Change list
- CustomCharacterService: rename model variable to customCharacterModel
- CustomCharacterService: validate if CustomCharacter is already exists. Return with 409 error code if yes.
- Move the CustomCharacter request validations to a separate method

### DIFF
--- a/src/__tests__/player/customCharacter/CustomCharacterService/createOne.test.ts
+++ b/src/__tests__/player/customCharacter/CustomCharacterService/createOne.test.ts
@@ -37,11 +37,11 @@ describe('CustomCharacterService.createOne() test suite', () => {
         });
     });
 
-    it('Should return created custom character if the input is valid and should return NOT_UNIQUE if already exists with this characterId', async () => {
+    it('Should return created custom character if the input is valid', async () => {
         const characterId = CharacterId.Prankster_202;
         const characterToCreate = createCustomCharacterBuilder.setCharacterId(characterId).build();
 
-        let [createdCharacter, errors] = await characterService.createOne(characterToCreate, player._id);
+        const [createdCharacter, errors] = await characterService.createOne(characterToCreate, player._id);
 
         const expectedSpecs = CharacterBaseStats[characterId];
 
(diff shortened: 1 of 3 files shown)


MikhailDeriabin (changes requested):
Good work, nice first PR!

Everything seems to be working, but I have found couple small things to change:

1. _customCharacter.service.ts_:

  - [x] [Line 50] Please use array destructuring here. [Here is more](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring) about array destructuring.
  ```ts
  const [isCharacterValid, characterValidationErrors] = await this.validateCustomCharacter(customCharacterToCreate, owner_id);
  if(characterValidationErrors) {
      ...
  ```

  - [x] [Line 51] and [Line 88] In the JS the convention about code style is a bit different from C#. 
  So here you should put the curly brace to the same line with the if-statement. 
  Jeri should configure the linter at some point, but right now you can use SonarCube plugin for your IDE as a linter
  [Here is more about code style in JS](http://www.w3schools.com/JS/js_conventions.asp).

  - [x] [Line 90] I think the better value of "value" field here is the `characterId` not `owner_id`.

  - [ ] [Line 95] The `IServiceReturn` defines the structure of the return statements. 
  So here you should return an array (tuple) consisting of two elements: on the first place data and on the second errors.
  But they can also be `null`. So that in case if there were no errors, you get `null` as a second element and in case of errors the first element is `null`.
  So here you most likely want to return:
  ```ts
  return [true, null];
  ```
  As well as change the return type of the `validateCustomCharacter()` method to:
  ```ts
  ... : Promise<IServiceReturn<true>> //where true defines the first elem type in return statement
  ```

 2. _tests/player/customCharacter/CustomCharacterService/createOne_
  
  - [ ] Can you a test that it is not possible to create two of the same type characters

PlayJeri:
@jkrizsan 

`npx jest path/to/test-file.test.ts`

jkrizsan:
image

do you have any idea? is it possible to run the tests locally?

jkrizsan:
@PlayJeri @MikhailDeriabin

Thanks for your recommendations. I can run the tests with the "Jest" plugin from UI. It seems there are in-memory tests.
I just extended one of the existing TC with this part. Originally I tried to implement a new TC, but it was able to run to green just alone, not together with the other TCs.

I know it is NOT the best solution. Is it OK?

MikhailDeriabin (approved):
Idk, I was able to move the test into a separate TC.

Anyway now everything is done, good work!

PR #482, opened 2025-03-30
Added logic and endpoints for tracking the number of online players.

PlayJeri:
This pull request introduces a new module for tracking the number of online players in the application. There are two new api endpoints, one for pinging the server and one to get the amount of online players.

The tracking works with redis. The ping endpoint sets a redis key with logged in player id for 5min TTL and the get endpoint counts the number of keys of online players.

### New Online Players Module:

* **Controller:**
  * Created `OnlinePlayersController` with endpoints to add a player to the online list and retrieve all online players. (`src/onlinePlayers/onlinePlayers.controller.ts`)
  * New endpoint `POST /online-players/ping`
  * New endpoint `GET /online-players`

* **Service:**
  * Implemented `OnlinePlayersService` to handle caching of online players and retrieval of the online players count. (`src/onlinePlayers/onlinePlayers.service.ts`)


MikhailDeriabin (changes requested):
Everything works nice.

Couple of things to do:

- [ ] Add also an endpoint where the online players list can be retrieved, I thing the player _ids will be enough for now.
- [ ] Add tests for the `OnlinePlayersService` class methods. I think you can use the actual `Cache` class in those. I did a little research and find out that the Nest is using the "keyv" package for the `Cache` and this package is using a standard JS Date object for [set](https://github.com/jaredwray/keyv/blob/main/packages/keyv/src/generic-store.ts#L129) and for [get](https://github.com/jaredwray/keyv/blob/main/packages/keyv/src/generic-store.ts#L113) methods. So that basically means that you can mock the JS Date object's `now()` method to return the appropriate values and also check that the cache gets invalidated = the player is no longer online. By appropriate values I mean for example to set the return value of `now()` method to 0 before adding player to the cache, and then returning some value which is after the TTL time (i.e. 400*1000), and then check that the `getAllOnlinePlayers()` is returning zero players. This is not really a good practice, but better than nothing.
- [ ] Consider adding some central storage for the cache keys, so that there are no name clashes in the future. The `LeaderboardService` seems to be using some own keys too.

MikhailDeriabin:
@PlayJeri 

By central storage I meant some object or enum with cache keys prefixes / names such as `ONLINE_PLAYERS_KEY` in `OnlinePlayersService` class or `'clanLeaderboard'` in `LeaderboardService` class.

PlayJeri:
@MikhailDeriabin 

I see. That's a good idea I'll make the changes 👍

MikhailDeriabin:
@PlayJeri 

I had a look at the tests. It seems like it is not possible to just mock the `Date.now()`, at least not for the `cacheService.store.keys` call, because it still returns the array with invalidated keys. The reason why is that the API uses redis, and it removes the invalidated keys automatically, but the _keyv_ does not (which is the default package if nothing else is provided like here in tests). 

I have adjusted the tests so that they are not using the mock for the `Cache` class.

MikhailDeriabin:
@PlayJeri 

Will u add the enum for cache keys here or in another PR?

PlayJeri:
@MikhailDeriabin 

I can do it here. Should be done by tomorrow

MikhailDeriabin (approved):
nice work!

PR #507, opened 2025-04-08
Fixed flea market items using an old items schema

MikhailDeriabin:
### Brief description

The bug was caused when the methods for conversion between `FleaMarketItem` and `Item` in `FleaMarketHelperService` class  were trying to access non-existing fields from the  `FleaMarketItem` class. The bug was solved by adding `material` field to  `FleaMarketItem` class and according DTOs and by renaming `qualityLevel` to `rarity` inside the methods.

### Change list

- Update `FleaMarketItem` schema
- Update flea market items DTOs
- Fix causing errors methods


PlayJeri (commented):
@MikhailDeriabin 

I just read the docs for the nestjs mongoose and there they recommend using different syntax for the prop types.

`@Prop({ type: [String] })` would be the preferred way instead of the `@Prop({ type: Array<string> })`

I know the `Array<string>` is used in other schemas as well and isn't causing any issues but maybe we want to move to the recommended way?

Other than that everything looks good.

MikhailDeriabin:
@PlayJeri 

Sure, I have done the changes for the `Item` and `FleaMarketItem`fields schemas to be `[ T ]`

Can you give a me link to this article?

PlayJeri:
> @PlayJeri
> 
> Sure, I have done the changes for the `Item` and `FleaMarketItem`fields schemas to be `[ T ]`
> 
> Can you give a me link to this article?

@MikhailDeriabin 

Under the model injection you can see the example. [https://docs.nestjs.com/techniques/mongodb](url)
And here under the Arrays section. [https://mongoosejs.com/docs/schematypes.html#arrays](url)

MikhailDeriabin:
Thanks

MikhailDeriabin:
@PlayJeri 

It is not mentioned in these articles that you have to use the `[String]` notation, but I have asked chat gpt, and it says that the mongoose does not recognize the `Array<string>`, instead it is converted to just `Array` with no type in it. 

So yes, we should move to this notation, I make a separate issue for that. Good thing that I have noticed it.

PlayJeri:
@MikhailDeriabin 

It isn't listed as requirement but both docs use the [T] in examples. That's what I meant.

PlayJeri:
@MikhailDeriabin 

I also noticed that the `extends Document` that I have used in the nested schemas is not the correct way. I think they are used in the playerStats and messages. So those should propably be fixed at some point.

MikhailDeriabin:
@PlayJeri 

ok, got it. For the nested documents I think you can as it is done for the avatar field in the player schema, but yes I can do a separate ticket for that as well.

Can u approve this PR, or is there something else to be changed?

PlayJeri (approved):

PR #511, opened 2025-04-09
Added roles field to clan data

MikhailDeriabin:
### Brief description

`roles` field is added to clan data containing clan roles data, as well as required enums and clan initialization roles (leader, elder and member).

### Change list

- Added schema for clan role
- Added enums `ClanBasicRight` and `ClanRoleType`
- Added `roles` field to clan schema and DTO
- Fixed failing tests after adding the `roles` field to clan schema

### DIFF
--- a/src/clan/role/ClanRole.schema.ts
+++ b/src/clan/role/ClanRole.schema.ts
@@ -1,0 +1,32 @@
+import { ClanBasicRight } from './enum/clanBasicRight.enum';
+import { ClanRoleType } from './enum/clanRoleType.enum';
+import { Prop, Schema, SchemaFactory } from '@nestjs/mongoose';
+import { ObjectId } from 'mongodb';
+
+/**
+ * Defines clan role structure
+ */
+@Schema({ toJSON: { virtuals: true }, toObject: { virtuals: true } })
+export class ClanRole {
+  /**
+   * Unique for clan name of the role
+   */
+  @Prop({ type: String, required: true })
+  name: string;
+
+  /**
+   * Type of the role
+   */
+  @Prop({ type: String, enum: ClanRoleType, required: true })
+  claRoleType: ClanRoleType;
+
+  /**
+   * Object with basic rights that the role has
+   */
+  @Prop({ type: Object, required: true })
+  rights: Partial<Record<ClanBasicRight, true>>;
+
+  _id: ObjectId;
+}
+
+export const ClanRoleSchema = SchemaFactory.createForClass(ClanRole);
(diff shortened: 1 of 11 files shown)


src/clan/role/dto/clanRole.dto.ts
PlayJeri, line 1: There is a typo in line 15. claRoleType should be clanRoleType

src/clan/role/ClanRole.schema.ts
PlayJeri, line 1: Fix typo on line 21 claRoleType to clanRoleType

src/clan/role/ClanRole.schema.ts
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * Type of the role
+   */
+  @Prop({ type: String, enum: ClanRoleType, required: true })
+  claRoleType: ClanRoleType;
+
+  /**
+   * Object with basic rights that the role has
+   */
+  @Prop({ type: Object, required: true })
```
PlayJeri, line 26: Mongoose wont validate the values passed here since the Prop type is Object. Maybe the rights could just be an array instead of the record so the enum validation could be used or add some kind of pre save validation or custom validation to the schema. Or do we trust that the service and DTOs catch the invalid data?

`@Prop({
  type: Map,
  of: Boolean,
  required: true,
  validate: {
    validator: (map: Map<string, boolean>) => {
      const allowedKeys = new Set(Object.values(ClanBasicRight));
      for (const [key, value] of map.entries()) {
        if (!allowedKeys.has(key)) return false;
        if (value !== true) return false;
      }
      return true;
    },
    message: 'Invalid rights map: keys must be valid ClanBasicRight and values must be true',
  },
})
rights: Map<ClanBasicRight, true>;
`

ChatGpt example of how to maybe do this.

src/clan/role/enum/clanBasicRight.enum.ts
```diff
@@ -0,0 +1,30 @@
...
+export enum ClanBasicRight {
+  /**
+   * Change items location in the soul home rooms
+   */
+  EDIT_SOULHOME = 'edit_soulhome',
+
+  /**
+   * Change clan data
+   */
+  EDIT_CLAN_DATA = 'edit_clan.data',
```
PlayJeri, line 13: Is the use of clan.data intentional? or should it be snake case edit_clan_data

PlayJeri (changes requested):
Looks good!

Fix the couple of typos and maybe think about the data structure of the "Role rights" Record vs Array and make sure the lack of validation in schema won't cause issues.

src/clan/role/enum/clanBasicRight.enum.ts
MikhailDeriabin: no, it is a typo

MikhailDeriabin:
@PlayJeri 

Everything is fixed. Role rights field should be record at least for now, since it is more flexible.

PR #513, opened 2025-04-10
Added ClanId guard

PlayJeri:
### Brief description
This PR introduces a new guard to validate that current user is a clan member.

### Change list

- Added optional `clan_id` field to the User class
- Added ClanIdGuard

### DIFF
--- a/src/__tests__/common/guard/clanIdGuard.test.ts
+++ b/src/__tests__/common/guard/clanIdGuard.test.ts
@@ -1,5 +1,4 @@
 import { ClanIdGuard } from '../../../common/guard/clanId.guard';
-import { ExecutionContext } from '@nestjs/common';
 import { Player, PlayerSchema } from '../../../player/schemas/player.schema';
 import { APIError } from '../../../common/controller/APIError';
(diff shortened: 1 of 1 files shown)


PlayJeri:
@MikhailDeriabin 

I think we talked about adding the clan id guard at some point but we never implemented it.
The new clan shop controller would benefit from having the clan id guard so I tought now might be a good time to add it.

This is the solution I came up with. It works on authenticated endpoint by getting the user data from request and then finding the user from db. If the player is not found or does not have clan_id field an error is thrown. If the clan ID is found it is attached to the request.

MikhailDeriabin (changes requested):
Looks good.

- [x] I would recommend not to use the `RequestHelperService` because it is kinda deprecated, use model injections instead.

- [x] Add tests for the `ClanIdGuard`. You can use builders for the `ExecutionContext` and `Request` from _test_utils/data_ folder
- [x] Add `clan_id` field to `UserBuilder` and `BoxUserBuilder` in _tests/auth_
- [x] Add `clan_id` field to `BoxUserProps` and `BoxUser` constructor in _box/auth_

MikhailDeriabin (approved):
Nice work

PR #514, opened 2025-04-10
Set player_id field to null by default for daily tasks

MikhailDeriabin:
### Brief description

Set the default value of `player_id` field in daily tasks to `null`, so that it could be easier to debug on the game side

### Change list

- Set the `player_id` field default value to `null`
- Set `player_id` to null for all daily tasks on server, which was not set
- Fix found bugs with `ClanRole` class


PlayJeri (approved):
Looks good.

PR #515, opened 2025-04-13
Added clan role information to player data

MikhailDeriabin:
### Brief description

Added clan member role information to the `Player` data as `clanRole_id`, so that it is possible to determine what role player has in his / her clan by the role _id. Since the player is not necessary gonna be in any clan, the default value of the `clanRole_id` is set to null.

Notice that the clan creator automatically will get the _Leader_ role in the created clan and that any new clan joined player will get the `Member` role.

Notice that it is not possible to set the `clanRole_id` field by making a request straight to the _/player_ POST or PUT endpoints and can be done only via a separate endpoints. 

### Change list

- Added `clanRole_id` to player DTO and DB schema
- Added `clanRole_id` to player testing data builders
- Fixed bugs raised after the `roles` field was added to clan data with signing in methods
- Fixed some failing tests

### DIFF
--- a/src/__tests__/auth/AuthService/signIn.test.ts
+++ b/src/__tests__/auth/AuthService/signIn.test.ts
@@ -56,11 +56,7 @@ describe('AuthService.signIn() test suite', () => {
     const clanDBFinal = {
       ...clearedClan,
       _id: clearedClan._id.toString(),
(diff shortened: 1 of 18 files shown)


PlayJeri (approved):
Looks good!

PR #520, opened 2025-04-15
Added endpoint for clan named roles creation

MikhailDeriabin:
### Brief description

Added endpoint _/clan/role_ POST, which allows clan members to create a new role. Notice that the role name must be unique within a clan and the role rights must also be unique for each role in a clan.

### Change list

- Added _/clan/role_ POST endpoint
- Added `@HasClanRights()` decorator for authorization check
- Fix bug with clan authorization rules throwing an error (due to use of `RequestHelperService`)
- Make `rights` field in the `ClanRole` schema to be an object to easier use in the code instead of `Map` class. The validation for this field is still in place.


src/clan/role/clanRole.service.ts
```diff
@@ -0,0 +1,97 @@
...
+    const [, clanUpdateErrors] = await this.basicService.updateOneById(
+      clan_id.toString(),
+      {
+        $push: { roles: [newRole] },
+      },
+    );
+
+    if (clanUpdateErrors) return [null, clanUpdateErrors];
+
+    const [updatedClan] = await this.basicService.readOneById<Clan>(
```
PlayJeri, line 88: Why is the clan fetched here? Isn't the roleToCreate dto valid so return the `[roleToCreate, null]`

src/clan/role/ClanRole.schema.ts
PlayJeri, line 1: The `CreateClanRoleDto` has validation to for name max length 50. Should that constraint be added to the schema?

PlayJeri (changes requested):
Looks good.

- [x] Add the name max length validation to clan role schema.
- [x] Rename the isRoleDuplicate to be more accurate
- [ ] Make sure the clan fetching in ClanRoleService createOne is actually necessary.

src/clan/role/clanRole.service.ts
MikhailDeriabin: It is fetched here because of the role `_id` field, which the `roleToCreate` does not have

PlayJeri (approved):

PR #531, opened 2025-04-25
Added /clan/role/set PUT endpoint for setting a role for a clan member

MikhailDeriabin:
### Brief description

Added _/clan/role/set_ PUT endpoint to set a default or named role to a specified player. Notice the limitations of this endpoint in the attached issue.

### Change list

- Added _/clan/role/set_ PUT endpoint
- Added tests
- Fixed bug with `APIError` error filter by introducing a new wrapper class, which holds an array of `APIError` and behaves similarly to array


PlayJeri (approved):
Looks good.

PR #533, opened 2025-04-25
Fixed a bug where /clan GET endpoint does not return pagination data

jkrizsan:
### Brief description
OffsetPaginate validate if the response already contains the metaData and wont populate the paginationData into the response if the metaData is missing. UniformResponse populate the metaData, it need to be executed before than the OffsetPaginate.

### Change list

- clan.controller:  put the UniformResponse attribute on top of the OffsetPaginate attribute in order to get populated the paginationData object in the response for the get clan api


PlayJeri (approved):
Nice and simple. Good job!

MikhailDeriabin (approved):
Nice work

PR #534, opened 2025-04-26
Feature add clan logo to player leaderboard data

PlayJeri:
### Brief description
this PR enhances the player leaderboard data by adding the clan logo data to it and fixes the missing pagination data from response.

### Change list

- New DTO for leaderboard players that extends the PlayerDto with added clanLogo prop.
- Replaced PlayerDto with LeaderboardPlayerDto in `@UniformResponse` decorator in player leaderboard endpoint.
- Added logic to `LeaderboardService` to fetch clan logos for players.

### DIFF
--- a/src/leaderboard/leaderboard.service.ts
+++ b/src/leaderboard/leaderboard.service.ts
@@ -5,8 +5,9 @@ import { IGetAllQuery } from '../common/interface/IGetAllQuery';
 import { CACHE_MANAGER, Cache } from '@nestjs/cache-manager';
 import ServiceError from '../common/service/basicService/ServiceError';
 import { SEReason } from '../common/service/basicService/SEReason';
-import mongoose from 'mongoose';
+import mongoose, { Model } from 'mongoose';
 import { CacheKeys } from '../common/enum/cacheKeys.enum';
+import { PlayerDocument } from 'src/player/schemas/player.schema';
 
 @Injectable()
 export class LeaderboardService {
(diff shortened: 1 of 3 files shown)


src/leaderboard/leaderboard.service.ts
```diff
@@ -67,7 +67,25 @@ export class LeaderboardService {
...
+          .exec();
+
+        const clanMap = clans.reduce((map, clan) => {
+          map[clan._id] = clan.clanLogo;
+          return map;
+        }, {});
+
+        data = fetchedData.map((player) => ({
+          ...player.toObject(),
+          clanLogo: clanMap[player.clan_id.toString()] || null,
```
MikhailDeriabin, line 84: This line throws an error if clan _id is null or undefined and `toString()` is called

src/leaderboard/leaderboard.service.ts
```diff
@@ -67,7 +67,25 @@ export class LeaderboardService {
     if (!data) {
       const fetchedData = await model.find().sort({ points: -1 }).exec();
       if (!fetchedData) throw new ServiceError({ reason: SEReason.NOT_FOUND });
-      data = fetchedData;
+
+      if (model === this.playerService.model) {
```
MikhailDeriabin, line 71: Consider to refactor the whole `if-else` block. 

For example, it can be a callback function for clan and player leaderboards data fetching.

MikhailDeriabin (changes requested):
Looks nice, but there are couple of things to be fixed

src/leaderboard/leaderboard.service.ts
PlayJeri: I refactored. the if-else block into two helper methods.

src/leaderboard/leaderboard.service.ts
PlayJeri: I removed the toString() call since it was unnecessary.

MikhailDeriabin (approved):
Nice work

PR #536, opened 2025-04-26
Added permission validation for edit clan data right for /clan PUT endpoint

MikhailDeriabin:
### Brief description

Added edit soul home permission validation for /clan PUT endpoint, which updates any clan data. Notice that logged-in player must also be in the updating clan.

### Change list

- Added edit soul home permission validation for _/clan_ PUT endpoint


PlayJeri (approved):
Look good!

PR #538, opened 2025-04-26
Fixed wrong response types retunred from endpoints

MikhailDeriabin:
### Brief description

Fixed endpoints, which was returning 200 status code instead of 204. The errors was caused by `Send204OnEmptyResInterceptor` class did not await the incoming data.

### Change list

- Fixed _/profile_ DELETE endpoint
- Fixed _/clan_ PUT endpoint
- Fixed _/clan_ DELETE endpoint
- Fixed _/item_ PUT endpoint


PlayJeri (approved):
Looks good

PR #541, opened 2025-04-30
Change player battle characters storing mechanism   517

jkrizsan:
### Brief description

Change validation of Player.battleCharacter_ids to also accept null as a value.

### Change list

- Add @IsMongoIdOrNull() decorator
- Replace the battleCharacter_ids @IsMongoId() decorator with @IsMongoIdOrNull() in the createPlayer/updatePlayer DTOs
- Made all player (DTO and schema) battleCharacter_ids nullable
- Extends tests to cover the changes

### DIFF
--- a/src/__tests__/common/decorator/validation/IsMongoIdOrNull/IsMongoIdOrNull.test.ts
+++ b/src/__tests__/common/decorator/validation/IsMongoIdOrNull/IsMongoIdOrNull.test.ts
@@ -1,18 +1,23 @@
 import 'reflect-metadata';
 import { validate } from 'class-validator';
 import { IsMongoIdOrNull } from '../../../../../common/decorator/validation/IsMongoIdOrNull.decorator';
-import { ObjectId } from 'mongodb';
 
 describe('@IsMongoIdOrNull() test suite', () => {
+  let dto: TestDto;
+  let objectId:  string ;
+  beforeEach(async () => {
+    dto = new TestDto();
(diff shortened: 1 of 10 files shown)


MikhailDeriabin (changes requested):
Good work. Pay attention to branch naming, it should not start with a capital letters and in the end there should be one dash not three.

Some changes need to be made also for the `battleCharacter_ids` field type:

- [x] change all player DTO `battleCharacter_ids` field type, that it also can be null
- [x] change all player schema `battleCharacter_ids` field type, that it also can be null
- [ ] change player builders setters to accept also null values (folder _/tests/player/data/player_)

jkrizsan:
@MikhailDeriabin

updated

PlayJeri (changes requested):
I think the best practice for validating arrays with class validator is to use `@IsArray()` and `@Constraint({ each: true })` togehter. This would also make the logic of the constraint more straight forward. The `@IsMongoIdOrNull` name doesn't really tell that it validates an array. 

With the each: true I think the validation logic could just be `return value === null || isMongoId(value);`

- [ ] Change constraint to only validate single value.
- [ ] Change DTO's to use `@IsMongoIdOrNull({ each: true )}`

jkrizsan:
@PlayJeri  updated

PlayJeri (approved):
Good work!

MikhailDeriabin (approved):

PR #544, opened 2025-05-03
Added DB transactions support

MikhailDeriabin:
### Brief description

Add support for replica set in MongoDB.

It was required to update the mongo version, as well as driver for it in npm. Due to that all of the other packages needed to be updated as well to the most recent once due to compatibility issues.

After the update, I have done manual testing and found some bugs in different endpoints, mb they was there before or appeared after the updates. I fixed found bugs as well.

For some reason the cache manager library stopped to work after update, which was in use in online players and leaderboard modules. So I had to add a module for accessing the Redis DB and remove the cache manager libraries from package.json

### Change list

- Updated mongoDB version
- Updated all npm packages to the newest
- Added scripts for setting up mongo to work with replica sets for Windows and Linux
- Fixed found bugs
- Fixed failing tests


PlayJeri (approved):
Wow! That was a lot more work than I anticipated. Great job!!! 👍 I can start working to get the transactions working next sprint then.

PR #547, opened 2025-05-06
decorate all _ids containing fields in returning data DTOs with `@ExtractField()`

MikhailDeriabin:
### Brief description

The bug with _id fields returning values was caused due to improper serialization, which is fixed by decorating then with the `@ExtractField()` decorator.

### Change list

- Decorate all DTO _id containing fields with `@ExtractField()`


PlayJeri (approved):
Good work

PR #555, opened 2025-05-15
Added an endpoint to add coins to the clan

jkrizsan:
### Brief description

Add a endpoint to add coins to the clan

### Change list

- add shop module
- add clanCoins endpoint
- fix the failed tests
- add shop test module and tests for the shop module

### DIFF
--- a/src/clan/dto/updateClan.dto.ts
+++ b/src/clan/dto/updateClan.dto.ts
@@ -78,7 +78,4 @@ export class UpdateClanDto {
   @IsEnum(Language)
   @IsOptional()
   language?: Language;
-
-  @IsOptional()
-  gameCoins: number;
 }
--- a/src/clan/dto/updateClan.dto.ts
+++ b/src/clan/dto/updateClan.dto.ts
@@ -78,4 +78,7 @@ export class UpdateClanDto {
   @IsEnum(Language)
   @IsOptional()
   language?: Language;
+
+  @IsOptional()
+  gameCoins: number;
 }
--- a/src/shop/buy/clanCoins.service.ts
+++ b/src/shop/buy/clanCoins.service.ts
@@ -1,24 +1,22 @@
 import { Injectable } from '@nestjs/common';
 import { ClanService } from '../../clan/clan.service';
 import { ClanCoinsDto } from './dto/clanCoins.dto';
-import { UpdateClanDto } from '../../clan/dto/updateClan.dto';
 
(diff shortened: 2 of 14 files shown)


src/clan/dto/updateClan.dto.ts
```diff
@@ -78,4 +78,7 @@ export class UpdateClanDto {
   @IsEnum(Language)
   @IsOptional()
   language?: Language;
+
```
MikhailDeriabin, line 81: This field can not be here, because that means that the gameCoins can be updated straight from the /clan PUT endpoint.

jkrizsan: Indeed, that's the job of the /clan PUT endpoint to update the clan properties :)
Anyway, I dropped the extra feature...

src/shop/buy/clanCoins.service.ts
```diff
@@ -0,0 +1,22 @@
...
+  ) {}
+    
+  async addCoins(
+    body: ClanCoinsDto
+  ) {
+    const [clan, error] = await this.clanService.readOneById(body.clan_id);
+    if (error) return [null, error];
+    
+    const gameCoins = clan.gameCoins + body.amount;
+    const { _id, } = clan;
```
PlayJeri, line 18: This can be done with just one db operation. You can remove the first read and just update the gameCoins using the $inc operator.

src/shop/buy/dto/clanCoins.dto.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Coins } from '../../../shop/enum/coins.enum.dto';
+import { IsClanExists } from '../../../clan/decorator/validation/IsClanExists.decorator';
+
+@AddType('ClanCoinsDto')
+export class ClanCoinsDto {
+  
+  @IsClanExists()
+  @IsMongoId()
+  @IsOptional()
+  clan_id: string;
```
PlayJeri, line 16: The clan_id gets overwritten by the LoggedUser clan id so this prop is not needed

PlayJeri (changes requested):
I think the DTOs shouldn't ask for information that gets overwritten. So consider removing the clan id from dto and use different params for the service method. And in the service method it is unnecessary to read the clan from db. You can directly increment the coins using the $inc operator.

PlayJeri (approved):
Nice work! 👍

PR #565, opened 2025-05-20
Added "status" field for online players

MikhailDeriabin:
### Brief description

Online players now have a `status` field, which determines what player is doing or where he/she is in the game. The possible values of it are:
 - UI (player just online somewhere in UI)
 - BattleWait (player is waiting for a battle)
 - Battle (player is playing the battle)

Notice that if the status value is not specified in the _/online-players/ping POST_, the UI value is set by default

### Change list

- Added body with optional field `status` for _/online-players/ping POST_ endpoint
- Added `status` field in return of _online-players GET_ endpoint
- Refactored `RedisService` and the use of it in the `onlinePlayers` module


PlayJeri (approved):
Good job

PR #566, opened 2025-05-20
Added filtering for /online-players GET enpoint by `status` field

MikhailDeriabin:
### Brief description

Now it is possible to filter online players by `status` field. It is possible to filter them by one or by multiple values. For example:

`?search=status="UI" //only players with status UI` 
or 
`?search=status="UI";AND;status="Battle" //only players with status UI or Battle`

### Change list

- Added search query to _/online-players GET_ endpoint


PlayJeri (approved):
Good job

PR #571, opened 2025-06-08
Added voting for clan role assigning to a clan member

PlayJeri:
### Brief description
This PR includes changes to the voting related functionality and adds the new clan role assigning voting. Important thing to notice is a breaking change for notifier MQTT. Old notifications had `item: dto` new one has `entity: dto`. This is because the entity to be voted on isn't necessarily an item anymore but it can be the role assigning related data.

### Change list

- Refactored voting-related enums and terminology. Such as replacing `ItemVoteChoice` with `VoteChoice`.
- Updated `VotingNotifier` to use more accurate `entity` instead of `item`.
- Added separate queues to FleaMarket, ClanShop and ClanRoles
- Fixed tests
- Added tests for new logic


MikhailDeriabin:
Looks good.

Can you update the swagger comments that now the voting is required to update clan role of the player, it should the _/clan/role/set PUT_ endpoint in `ClanRoleController`. Also check other endpoints swagger that might be affected of the changes. 

Also, add the swagger tag for the next release as it is described [here](https://github.com/Alt-Org/Altzone-Server/wiki/In%E2%80%90code-swagger-descriptions#general)

MikhailDeriabin (approved):

PR #572, opened 2025-06-14
Added player guest account functionality

jkrizsan:
### Brief description

Add /profile/guest POST endpoint.
Add tests for the new ProfileService method

### Change list

- Add isGuest field to Profile schema
- Moved src/box/tester/passwordGenerator.ts to under the common folder
- Add mechanism to generate a new guest account
- Add /profile/guest POST endpoint.
- Add tests for the new ProfileService method


src/profile/profile.controller.ts
```diff
@@ -86,6 +88,21 @@ export default class ProfileController {
     return [createdProfile, errors];
   }
 
+  @NoAuth()
```
MikhailDeriabin, line 91: Add swagger docs for this endpoint, remember about the swagger release tag, which marks this endpoint as a new one (you most likely want to get this tag from the dev branch):

- [About swagger docs in code](https://github.com/Alt-Org/Altzone-Server/wiki/In%E2%80%90code-swagger-descriptions)

src/profile/profile.service.ts
```diff
@@ -65,6 +67,32 @@ export class ProfileService
     });
   }
 
+  async createGuestAccount(
```
MikhailDeriabin: Add JSDocs for the method

src/profile/profile.service.ts
```diff
@@ -65,6 +67,32 @@ export class ProfileService
...
+  ): Promise<object> {
+    const password = this.passwordGenerator.generatePassword('fi');
+    const username = "guest-account-" + password;
```
MikhailDeriabin: I think you should not expose the password like that, instead of password add `new ObjectId().toString()` here. 

Also, it is not necessary that this way you will not get duplicates, since the password generator does not guarantee a unique values.

src/profile/profile.service.ts
```diff
@@ -65,6 +67,32 @@ export class ProfileService
     });
   }
 
+  async createGuestAccount(
+  ): Promise<object> {
```
MikhailDeriabin: Consider adding a proper type here, you should avoid `object` types, where you can.

It seems like it can be of type IServiceReturn

src/profile/profile.module.ts
```diff
@@ -7,17 +7,38 @@ import { RequestHelperModule } from '../requestHelper/requestHelper.module';
...
+import { AuthService } from '../auth/auth.service';
+import { AuthServiceProvider } from '../auth/authService.provider';
+import { ClanSchema } from '../clan/clan.schema';
+import { PlayerSchema } from '../player/schemas/player.schema';
+import BoxAuthService from '../auth/box/BoxAuthService';
+import { BoxSchema } from '../box/schemas/box.schema';
+import { GroupAdminSchema } from '../box/groupAdmin/groupAdmin.schema';
+import { AuthModule } from '../auth/auth.module';
 
 @Module({
```
MikhailDeriabin, line 20: I have adjusted imports and providers sections. Please read the NestJS docs about [modules](https://docs.nestjs.com/modules). 

Please have a look at the commented out code, I did not remove it on purpose, so it is easier to analyze for you. When you are done, remove the commented out code.

src/profile/profile.module.ts
```diff
@@ -7,17 +7,38 @@ import { RequestHelperModule } from '../requestHelper/requestHelper.module';
...
   imports: [
     MongooseModule.forFeature([
       { name: ModelName.PROFILE, schema: ProfileSchema },
+      // { name: ModelName.CLAN, schema: ClanSchema },
```
MikhailDeriabin, line 24: Usually you do not need to import other modules schemas. Sometimes you need it it tests

src/profile/profile.module.ts
```diff
@@ -7,17 +7,38 @@ import { RequestHelperModule } from '../requestHelper/requestHelper.module';
...
+      // { name: ModelName.PLAYER, schema: PlayerSchema },
+      // { name: ModelName.BOX, schema: BoxSchema },
+      // { name: ModelName.GROUP_ADMIN, schema: GroupAdminSchema },
     ]),
     PlayerModule,
     RequestHelperModule,
+    AuthModule,
```
MikhailDeriabin, line 31: If you use something from this module, such as `AuthService`, which is exported from it, just import the module

src/profile/profile.module.ts
```diff
@@ -7,17 +7,38 @@ import { RequestHelperModule } from '../requestHelper/requestHelper.module';
...
   ],
   controllers: [ProfileController],
-  providers: [ProfileService, isProfileExists],
+  providers: [
+    ProfileService,
+    isProfileExists,
+    PasswordGenerator,
```
MikhailDeriabin, line 37: This one is work around, because the `PasswordGenerator` does not have own module, so it is ok here

src/profile/profile.module.ts
```diff
@@ -7,17 +7,38 @@ import { RequestHelperModule } from '../requestHelper/requestHelper.module';
...
+    // AuthService,
```
MikhailDeriabin, line 38: It is not a provider of this module, so it is incorrect use of the providers array

MikhailDeriabin (changes requested):

MikhailDeriabin (approved):

PR #594, opened 2025-07-02
Changed testing box creation logic

MikhailDeriabin:
### Brief description

Now when a box is created only testing group admin profile and player are created and clans are not.

### Change list

- Remove clan creation logic from box creation
- Adjust some other classes and tests to not contain any clan logic creation
- Adjust documentation to not mention any logic related to clans creation


PlayJeri (approved):
Good job

PR #596, opened 2025-07-06
Change claiming tester account logic

MikhailDeriabin:
### Brief description

The claiming account logic is now working without need of cookies. The endpoint itself is changed from _/box/claim-account GET_ to _/box/claim-account POST_ (due to overlapping with _/box/:_id GET_ endpoint).

Also according to plan the new profile and player are created, and the player is assigned to some of the box clans, during the account claiming process. Unlike it was before, when the profiles and players were already created before the claiming account step.

New `SEReason` `NOT_AUTHORIZED` was added.

### Change list

- Account claiming logic is changed
- Account claiming logic is moved to a separate inner module in box
- `TesterAccount` service is also moved with account claiming logic
- Some payload and DTO classes are changed to be more suitable for the new logic, but overall API interface did not change
- New `SEReason` `NOT_AUTHORIZED` was added
- box module was refactored
- tests for claiming account were added as well as some old tests were adjusted to not fail with the new logic
- JSDocs were updated for account claiming logic related methods


src/box/accountClaimer/accountClaimer.service.ts
PlayJeri, line 1: I don't know if you considered using transaction in the claimAccount method. Since the box data gets deleted after testing it propably isn't 100% necessary but using transaction when updating multiple documents that depend on each other is recommended.

PlayJeri (approved):
Use transactions when making changes to multiple documents that effect each other from the same function call. Risks in this testing environment is minimal but just as a reminder when updating the production code.

PR #597, opened 2025-07-07
Added `box_id` field to every schema in DB for TESTING_SESSION environment

MikhailDeriabin:
### Brief description

The field is added via mongoose plugin. For now, it is very strict and has a default value of null just in case. Later it can be adjusted to be a required field, when everything is ready and can be properly tested.

Also, indexing is added to this field, because it is expected to be used in every fetch query in the TESTING_SESSION environment.

Notice that this branch is made from the _change/claiming-account-functionality-585_ branch wherefore the PR contains changes from that branch as well. So it is better to have a look at #596  PR first 

### Change list

- Added plugin, which adds `box_id` field to all collections


PlayJeri (approved):
Good job

PR #606, opened 2025-07-19
Added group admin creation endpoint

MikhailDeriabin:
### Brief description

Added _/box/createAdmin POST_ endpoint for creating group admins. The endpoint is temporary and was required for the teams building the client side to test the box functionality without a need to request to insert the admin manually to DB.

### Change list

- Added _/box/createAdmin POST_ endpoint
- Updated README to include the API version for teacher pages
- Added tests


src/box/box.controller.ts
```diff
@@ -180,55 +183,75 @@ export class BoxController {
   }
 
   /**
-   * Get box by _id, For time of development only
+   * Create a group admin. The endpoint for time of development only
+   *
+   * @remarks Endpoint for getting box data by its _id.
```
PlayJeri, line 188: The remarks seems to be from `getOne` endpoint.

src/box/box.controller.ts
```diff
@@ -180,55 +183,75 @@ export class BoxController {
...
+   *
+   * The group admin is required in order to use the endpoints starting with "/box",
+   * i.e. for box initialization or changing its settings. Read the docs for more info.
+   */
+  @ApiResponseDescription({
+    success: {
+      status: 204,
+    },
+    errors: [409],
```
PlayJeri, line 197: Is there a possibility of other errors like 400 in case of invalid body?

src/box/box.controller.ts
```diff
@@ -180,55 +183,75 @@ export class BoxController {
...
+  @UniformResponse()
+  public async createAdmin(@Body() body: CreateGroupAdminDto) {
+    const [, errors] = await this.groupAdminService.createOne(body);
+    if (errors) return [null, errors];
+  }
+
+  /**
+   * Get all boxes. The endpoint for time of development only
    *
    * @remarks Endpoint for getting box data by its _id
```
PlayJeri, line 212: Remarks seem to be from get one by id not get all.

src/box/groupAdmin/groupAdmin.service.ts
PlayJeri, line 1: the `validateAdmin` and explicit `createOne` methods might be unnecessary. I think you could just use the basic service here. Since this method is only called from the create group admin endpoint and there is already an DTO class to validate the input so you could just call the basic service without these methods. If you want more explicit error messages those could be added staright to the DTO

PlayJeri (changes requested):
Check the remarks and make sure that the new methods in the group admin service are actually required.

MikhailDeriabin:
@PlayJeri 

Everything is fixed

PR #609, opened 2025-07-24
Added chats moderation in testing sessions

jkrizsan:
### Brief description

Add moderation for testing session chats

### Change list

- Add changing a particular chat message content functionality
- Add removing a particular chat message content functionality
- Add tests

### DIFF
--- a/src/__tests__/chat/chatService/deleteChatMessageById.test.ts
+++ b/src/__tests__/chat/chatService/deleteChatMessageById.test.ts
@@ -13,28 +13,25 @@ describe('ChatService.deleteChatMessageById() test suite', () => {
   const chatModel = ChatModule.getChatModel();
   const chatMessageBuilder = ChatBuilderFactory.getBuilder('ChatMessage');
 
-  const clan1ID = new ObjectId().toString();
-  const player1ID = new ObjectId();
+  const clanID = new ObjectId().toString();
+  const playerID = new ObjectId();
   let clanChatToCreate1: ChatMessage;
 
   const clanMessages = [];
 
-  beforeAll(() => {
-    clanChatToCreate1 = chatMessageBuilder
-      .setType(ChatType.CLAN)
-      .setSenderId(player1ID)
-      .setClanId(clan1ID)
-      .build();
-
-    clanMessages.push(clanChatToCreate1);
-  });
-
   beforeEach(async () => {
     env.ENVIRONMENT = Environment.TESTING_SESSION;
     await chatModel.deleteMany({});
-    jest.clearAllMocks();
 
     chatService = await ChatModule.getChatService();
+
+    clanChatToCreate1 = chatMessageBuilder
+      .setType(ChatType.CLAN)
+      .setSenderId(playerID)
+      .setClanId(clanID)
+      .build();
+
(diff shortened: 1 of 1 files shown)


src/chat/chat.controller.ts
```diff
@@ -74,4 +86,47 @@ export class ChatController {
...
+    success: {
+      status: 204,
+    },
+    errors: [],
+    hasAuth: true,
+  })
+  @IsGroupAdmin()
+  @UniformResponse(ModelName.CHAT)
+  @Patch()
+  async configureBox(@Body() body: UpdateChatMessageDto) {
```
PlayJeri, line 106: Change the function name

src/chat/service/chat.service.ts
```diff
@@ -83,4 +88,75 @@ export class ChatService {
...
+   * - REQUIRED if _id is not provided
+   */
+  async updateOneById(
+    chat: Partial<UpdateChatMessageDto>,
+  ): Promise<[boolean | null, ServiceError[] | null]> {
+    if (env.ENVIRONMENT !== Environment.TESTING_SESSION) {
+      return await this.getmisconfiguredEnvironmentError();
+    }
+
+    if (!chat._id)
```
PlayJeri, line 108: The _id field is required in the update dto so this if statement can be removed

src/common/enum/modelName.enum.ts
```diff
@@ -18,6 +18,7 @@
  */
 export enum ModelName {
   CHARACTER_CLASS = 'CharacterClass',
+  CHAT = 'Chat',
```
PlayJeri, line 21: The CHAT_MESSAGE is used for the model since its more descriptive. The CHAT model doesn't exist so this can be removed.

PlayJeri (changes requested):

src/chat/service/chat.service.ts
```diff
@@ -83,4 +88,75 @@ export class ChatService {
...
+   * @param chat - The data needs to be updated of the ChatMessage.
+   * @returns _true_ if ChatMessage was updated successfully, _false_ if nothing was updated for the ChatMessage,
+   * or a ServiceError:
+   * - NOT_FOUND if the ChatMessage was not found
+   * - REQUIRED if _id is not provided
+   */
+  async updateOneById(
+    chat: Partial<UpdateChatMessageDto>,
+  ): Promise<[boolean | null, ServiceError[] | null]> {
+    if (env.ENVIRONMENT !== Environment.TESTING_SESSION) {
```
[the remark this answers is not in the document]
jkrizsan: ok, however controller classes are not covered with tests..

src/common/enum/modelName.enum.ts
jkrizsan: ok, in this case the controller name misleading

PlayJeri:
Are we okay adding the forward refs back in to the api. Are they necessary here and could they be avoided if the message updating endpoints would be in the box module? Since the message updating is only testing and box related it might ge a good idea? If you are fine with this current implementation so am I but just wanted to suggest this.

PlayJeri (approved):

PR #619, opened 2025-08-02
Added clan stalls for flea market items

jkrizsan:
### Brief description

Add a way to get a list of all stalls

Add a way to get a stall by clan_id with its items

### Change list

- add stall to the clan schema
- add stall to the Clan Dtos
- add stall submodule to the FleaMarket module
- Add getAll stall API
- Add getStall by ClanId API
- Add tests for Stall module


src/fleaMarket/stall/stall.service.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Returns the stall by clan id
+   */
+  async readOneByClanId(
+    clanId: string,
+  ): Promise<[StallResponse | null, ServiceError[] | null]> {
+    const [clan, error] = await this.clanService.readOneById(clanId);
+
+    if (!clan || !clan.stall || error) {
```
PlayJeri, line 19: The `clanService.readOneById` uses the basic service method that already returns an specific error if the clan is not found so you could do something like this.
```suggestion
    if (error) {
		return [null, error]
	}
```

In case the stall wouldn't be found it would return a Service Error with NOT_FOUND reason.

src/fleaMarket/stall/stall.service.ts
```diff
@@ -0,0 +1,59 @@
...
+      ];
+    }
+
+    return [clan.stall, null];
+  }
+
+  /**
+   * Returns all stalls for all clans
+   */
+  async readAll(): Promise<[StallResponse[] | null, ServiceError[] | null]> {
```
PlayJeri, line 37: Use the `IServiceReturn` type in the explicit return types.

src/fleaMarket/stall/stall.service.ts
```diff
@@ -0,0 +1,59 @@
...
+          new ServiceError({
+            reason: SEReason.NOT_FOUND,
+            message: 'Failed to retrieve clans with stalls.',
+          }),
+        ],
+      ];
+    }
+
+    return [
+      clans.filter((clan) => !!clan.stall).map((clan) => clan.stall),
```
PlayJeri, line 55: Is the filter here needed since the readAll call already has the `filter: { stall: { %ne: null } }` filter.

src/fleaMarket/stall/stall.service.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Returns all stalls for all clans
+   */
+  async readAll(): Promise<[StallResponse[] | null, ServiceError[] | null]> {
+    const [clans, error] = await this.clanService.readAll({
+      filter: { stall: { $ne: null } },
+    });
+
+    if (!clans || clans.length === 0 || error) {
```
PlayJeri, line 42: Same here. The `clanService.readAll` uses the basic service and will return a specific error so no need to define the new Service error here.

PlayJeri (changes requested):

src/fleaMarket/stall/payloads/stallResponse.ts
```diff
@@ -0,0 +1,44 @@
+import { Expose } from 'class-transformer';
+import AddType from '../../../common/base/decorator/AddType.decorator';
+
+/**
+ * Represents the advertisement poster for a stall
+ */
+@AddType('AdPoster')
+export class AdPoster {
```
MikhailDeriabin, line 8: Most likely you do need to export it, if you use it only here

src/fleaMarket/stall/payloads/stallResponse.ts
```diff
@@ -0,0 +1,44 @@
...
+   * Main furniture used as the focal point of the stall's poster
+   */
+  @Expose()
+  mainFurniture?: string;
+}
+
+/**
+ * Represents a Clan's stall in the flea market
+ */
+@AddType('StallResponse')
```
MikhailDeriabin, line 31: It should be in _dto_ folder.

The file name should end with `.dto.ts` (for swagger generation).

Also add examples for swagger here.

If you want to leave the payload class as well, you can copy this one and remove the `@Expose()` decorators

src/fleaMarket/stall/stall.service.ts
MikhailDeriabin: Yes, the `IServiceReturn` type is a tuple, which always follows the logic:

If everything is ok, the first element in tuple will be the response and second will be `null`

If error occurred, the first element will be `null` and the second array with errors.

MikhailDeriabin (changes requested):
Overall good work

PlayJeri (approved):
Goob job.

MikhailDeriabin (approved):

PR #620, opened 2025-08-02
Change removing testing session logic

MikhailDeriabin:
### Brief description

Change _/box DELETE_ endpoint, which now removes all box data based on the `box_id` field.

### Change list

- Change box deletion logic
- Fix tests for box deletion and box scheduler

### DIFF
--- a/src/box/box.scheduler.ts
+++ b/src/box/box.scheduler.ts
@@ -26,6 +26,7 @@ export class BoxScheduler {
     for (const box of boxes) {
       if (box.boxRemovalTime <= currentTime.getTime()) {
         await this.boxService.deleteBox(box._id.toString());
+        continue;
       }
 
       if (box.sessionResetTime <= currentTime.getTime()) {


src/box/box.controller.ts
```diff
@@ -36,6 +36,7 @@ import { GroupAdmin } from './groupAdmin/groupAdmin.schema';
 import { Model } from 'mongoose';
 import BasicService from '../common/service/basicService/BasicService';
 import { NoBoxIdFilter } from './auth/decorator/NoBoxIdFilter.decorator';
+import { errors } from 'mongodb-memory-server';
```
PlayJeri, line 39: This seems to be unused

src/box/box.scheduler.ts
```diff
@@ -34,8 +26,8 @@ export class BoxScheduler {
     for (const box of boxes) {
       if (box.boxRemovalTime <= currentTime.getTime()) {
         await this.boxService.deleteBox(box._id.toString());
-        continue;
```
PlayJeri, line 37: Why is the continue removed? If the box gets deleted the boxService.reset should not try to run.

PlayJeri (changes requested):

src/box/box.scheduler.ts
MikhailDeriabin: Right, I did not get why `continue` was here first

PlayJeri (approved):

PR #624, opened 2025-08-15
Added a selling price for selling items votings

jkrizsan:
### Brief description

Add price for selling item voting

### Change list

- Add the SellFleaMarketItemDto (item id, price)
- Use the new Dto with the handleSellItem method
- handleSellItem: update the fleaMarketItem price after voting was started
- Update the relevant tests


PlayJeri:
Doesn't this cause an issue where users can change the item price with this endpoint? You could start the voting with new price of 999999 and that will be updated as item price no matter the outcome of the voting and since the old value is replaced with the new one it can't be changed back. Also like I asked earlier in ossue comments why is this price field even necessary? Why isn't the item sold at the price that is already present for all items. And the price changing would be handled with the change item price voting?

@MikhailDeriabin 
@jkrizsan

PlayJeri:
I mean you could vote on the price before you put the item on sale for fleamarket

MikhailDeriabin:
@PlayJeri 

Ok, I see now. 

It is fine to have any price in the flea market. It is up to clan members, but problems can happen when the price of item in the stock can be set by the members. 

How about changing the `handleRejectedSellVoting()` line 394, to set the default price of an item from the _clanInventory/item/const/itemProperties.ts_ based on the item name?

PlayJeri:
That works but if there is a changed price already from passed voting that would then get overwritten. Maybe the fleamarket item voting schema could hold a sell price value that is used for successful votings? 

@MikhailDeriabin

PR #625, opened 2025-08-16
Refactored voting schemas

PlayJeri:
### Brief description

Refactored voting schemas. Now we have a base voting schema that has all the shared properties and then by using discriminators and "type" field as discriminatorKey explicit schemas for all the different types of votings. 

### Change list

- Refactored voting schemas
- Changed service methods so that they support the new schemas
- Added data builders for the new types
- Fixed tests


MikhailDeriabin:
@PlayJeri 

Looks good. 

I did not get why have you removed the "type" field from all voting schemas. I think it can be useful on the client side to have it, so mb add this to voting DTO?

PlayJeri:
The type has to be removed from the discriminator schemas since it gets added in when the discriminator is registered. So the type is present on all schemas in runtime. In the module.ts where the discriminators are registered the value field is the type. I don't know why it has to be this way but if you dont remove the type from the schemas the schema registering throws an error about the type field

MikhailDeriabin (approved):

PR #627, opened 2025-08-17
Fixed bug with flea market selling endpoint can change the stock item price

MikhailDeriabin:
### Brief description

Fixed bug with flea market selling endpoint can change the stock item price. Now item price is set only when the item is approved to be sold in the flea market.

### Change list

- Changed price setting logic
- Fixed tests


PlayJeri (approved):
Well done

PR #628, opened 2025-08-19
Added updating clan stall ad post functionality

jkrizsan:
### Brief description

As a clan member with right "Shop", I want to be able to change ad poster of my clan stall, so that the stall stays appealing for the shoppers.

### Change list

- Add Patch API for the Stall Controller to update the AdPoster
- Add updateAdPosterByClanId method to the StallService which one can manage the clans AdPoster update
- Add tests


MikhailDeriabin (approved):
Good job

PR #655, opened 2025-08-29
Change/item-voting-logic-614

PlayJeri:
### Brief description

Changed item selling logic on fleamarket items. Now when a sell fleamarket item vote passes the item is moved to fleamarket but its' status will be shipping not available. The new `POST /change-item-status` endpoint can be used to change the status to make items available or unavailable for other clans.

### Change list

- Add `POST /change-item-status` endpoint.
- Add getFleaMarketItemClanId helper method for validation
- Add checkClanItemSlots method for available slot validation
- Chang default status of fleamarket item from available to shipping.
- Add tests for the service methods.
- Add change item status dto


jkrizsan (approved):
Looks Good

PR #656, opened 2025-08-30
Feature/add player battle points 631

jkrizsan:
### Brief description
As a player, I want to get points when I win a battle, so that I can rise in a battle leaderboard.

### Change list

- Add battlePoints to the player schema and the DTOs
- Delete obsolete PlayerEvents(BATTLE_PLAYED, MESSAGE_SENT and VOTE_MADE)
- Delete the related obsolete codes as well
- Add BATTLE_LOSE type to the PlayerEvents
- Use battlePoints instead of the points in the PlayerRewarder.rewardForPlayerEvent
- Add relevant tests to cover the new logic

### DIFF
--- a/src/__tests__/player/data/player/playerBuilder.ts
+++ b/src/__tests__/player/data/player/playerBuilder.ts
@@ -8,6 +8,7 @@ export default class PlayerBuilder {
     name: 'defaultPlayer',
     backpackCapacity: 10,
     points: 0,
+    battlePoints: 0,
     uniqueIdentifier: 'unique-id',
     above13: true,
     parentalAuth: true,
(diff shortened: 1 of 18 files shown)


PlayJeri (changes requested):
Good job!

The increasePlayerBattlePoints should maybe be renamed to update or change since it can also reduce the points. And the logic can be simplified a bit I think. Maybe something like this.

```typescript
    private async updatePlayerBattlePoints(
    player_id: string,
    battlePoints: number,
  ): Promise<IServiceReturn<boolean>> {
    const [player, errors] =
      await this.playerService.readOneById<PlayerDto>(player_id);
    if (errors) return [null, errors];

    battlePoints = Math.max(0, player.battlePoints + battlePoints);
    return await this.playerService.updateOneById(player_id, {
      $set: { battlePoints },
    });
  }
  ```
  
  It can also be divided into two separate 'increase' and 'decrease' methods and the first read operation can be skipped on increase like you had in your previous example.

PlayJeri (approved):
Good job

PR #657, opened 2025-09-05
Feature/add battle points for clans 630

jkrizsan:
### Brief description
As a clan member, I want to get points for my clan when I win a battle, so that my clan can rise in a battle leaderboard.

### Change list

- Add battlePoints to the clan schema and relevant DTOs
- Extend the ClanRewarder service with a logic that can update the clans battlePoints similary how its works for the Players battlePoints
- use the ClanRewarder from the ClanEventHandler and GameEventHandler
- Add tests to cover the new code


PlayJeri:
> @PlayJeri no needed the player / clan id change?

No need. I dont know what i was thinking earlier. My apologies 😅

PR #659, opened 2025-09-05
feat: add tests for moving flea market items to stock

PlayJeri:
### Brief description
Add new endpoint to move items from fleamarket to stock.

### Change list

- Add `POST /move-to-stock/:_id` endpoint
- Add service methods to handle the item moving
- Add tests for the service methods


src/fleaMarket/fleaMarket.service.ts
```diff
@@ -34,6 +34,8 @@ import { itemNotAuthorizedError } from './errors/itemNotAuthorized.error';
 import ServiceError from '../common/service/basicService/ServiceError';
 import { SEReason } from '../common/service/basicService/SEReason';
 import { maxSlotsReachedError } from './errors/maxSlotsReached.error';
+import { ItemBookedError } from './errors/itemBooked.error';
+import { CreateItemDto } from 'src/clanInventory/item/dto/createItem.dto';
```
jkrizsan, line 38: I think here could be better something like "../clanInventory/item/dto/createItem.dto"

PR #668, opened 2025-09-18
Feature/add-boxId-query-for-testing-session-588

PlayJeri:
### Brief description

Added box ID to the query filters for paginated endpoints in testing session to count the found data correctly.

### Change list

- Added a box_id filter in offsetPagination interceptor
- Changed player and profile getAll endpoints away from deprecated version.
- Added the JwtModule registration for test common modules

### DIFF
--- a/src/__tests__/box/auth/BoxIdFilterInterceptor.test.ts
+++ b/src/__tests__/box/auth/BoxIdFilterInterceptor.test.ts
@@ -86,28 +86,38 @@ describe('BoxIdFilterInterceptor class test suite', () => {
     expect(result).toEqual(data);
   });
 
-  it('should extract box_id from JWT if user is missing', async () => {
-    const boxId = 'jwtBoxId';
-    mockJwtService.verifyAsync.mockResolvedValueOnce({ box_id: boxId });
-    const data = [
-      { box_id: 'jwtBoxId', value: 1 },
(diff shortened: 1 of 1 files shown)


src/common/interceptor/request/offsetPagination.interceptor.ts
```diff
@@ -93,6 +101,16 @@ export function OffsetPaginate(
...
       );
     }
+
+    private attachBoxIdToRequest(request: Request) {
+      const token = request.headers.authorization?.split(' ')[1];
+      try {
+        const payload: BoxUser = this.jwtService.verify(token);
+        request['user'] = new BoxUser(payload);
+      } catch (e) {
+        void e;
```
jkrizsan, line 111: I do not expect that so often will execute this line, but in general not a good approach to hide an exception in this way. Why you don't log it or rethrow?

PlayJeri:
@jkrizsan 

I discovered a better way to do this whole thing so I changed the logic. This new approach should be simpler and have less code overall. 

Found this when trying to explain why I skipped the error handling in the jwt verification. Reason why I skipped was because the BoxIdFilter interceptor handled the same error already. And that lead me to changing the jwt verification from the pagination decorator to boxId interceptor. This way the jwt service registration is no longer needed for the tests.

jkrizsan (approved):
Looks Good

PR #669, opened 2025-09-18
change avatar type from Avatar to AvatarDto

PlayJeri:
### Brief description

Fixed a bug where the sender's avatar data wan't exposed in the Sender DTO.

### Change list

- Changed the type of the avatar in Sender Dto


jkrizsan (approved):
Looks Good!

PR #673, opened 2025-09-22
add session schema and collection. Move session related props from box to session.

PlayJeri:
### Brief description

Divide the box schema into box and session schemas.

### Change list

- Add session schema and collection
- Move session related data from box to session schema
- Add session model to modelName enum
- Add session model to box module

### DIFF
--- a/src/box/schemas/session.schema.ts
+++ b/src/box/schemas/session.schema.ts
@@ -8,7 +8,7 @@ import { defaultPredefinedDailyTasks } from '../dailyTask/defaultPredefinedDaily
 import { HydratedDocument, Types } from 'mongoose';
 import { ClanToCreate, ClanToCreateSchema } from './ClanToCreate.schema';
 import { BoxReference } from '../enum/BoxReference.enum';
-import { ModelName } from 'src/common/enum/modelName.enum';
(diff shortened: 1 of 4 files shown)


src/box/schemas/session.schema.ts
```diff
@@ -0,0 +1,113 @@
...
+import { SessionStage } from '../enum/SessionStage.enum';
+import {
+  PredefinedDailyTask,
+  PredefinedDailyTaskSchema,
+} from '../dailyTask/predefinedDailyTask.schema';
+import { defaultPredefinedDailyTasks } from '../dailyTask/defaultPredefinedDailyTasks';
+import { HydratedDocument, Types } from 'mongoose';
+import { ClanToCreate, ClanToCreateSchema } from './ClanToCreate.schema';
+import { BoxReference } from '../enum/BoxReference.enum';
+import { ModelName } from 'src/common/enum/modelName.enum';
```
jkrizsan, line 11: please use relative path here

PlayJeri: I need to be careful with these. I don't know why the auto imports use the absolute path sometimes.

good catch

PR #676, opened 2025-09-28
Feature/add create a playlist for the clan daily task 635

jkrizsan:
### Brief description
Add support for "change your avatar's clothes" daily task.

### Change list

- Emit a server event when the PUT player api updated the avatars clothes


src/player/player.controller.ts
```diff
@@ -117,7 +123,19 @@ export default class PlayerController {
   @Authorize({ action: Action.update, subject: UpdatePlayerDto })
   @BasicPUT(ModelName.PLAYER)
   public async update(@Body() body: UpdatePlayerDto) {
-    return this.service.updateOneById(body);
+    const [player, _] = await this.service.getPlayerById(body._id);
+    const playerUpdateResults = this.service.updateOneById(body);
```
PlayJeri, line 127: await this call

src/player/player.controller.ts
```diff
@@ -117,7 +123,19 @@ export default class PlayerController {
...
+
+    if (playerUpdateResults instanceof Promise && body.avatar.clothes) {
```
PlayJeri, line 129: this line can then be removed

src/player/player.controller.ts
```diff
@@ -117,7 +123,19 @@ export default class PlayerController {
...
+      if (player.avatar.clothes !== body.avatar.clothes) {
```
PlayJeri, line 130: This would throw a runtime error if player or avatar is null.

So change is to `player?.avatar?.clothes !== body?.avatar?.clothes`

PlayJeri (changes requested):
- [ ] Fix the optional properties in the if statement
- [ ] Fix the failing tests (maybe caused by dependency changes)

jkrizsan:
Still can't fixed the tests.

jkrizsan:
@PlayJeri I found a way to make the test green and keep the event emitter functionality. Pls let me know if you have ideas to make it better.

PlayJeri (approved):
It works and it seems to be similar to other common modules so I see no problem with this. Good job.

PR #686, opened 2025-10-02
Fix stall buySlot and addPoster endpoints

PlayJeri:
### Brief description

Endpoints were missing the DetermineClanId decorator so the clan_id field was null.

### Change list

- Added DetermineClanId decorator to the endpoints in controller


jkrizsan (approved):
Looks Good

PR #688, opened 2025-10-02
Add the Expose decorator to organizer dto

PlayJeri:
### Brief description

Organizer data wasn't showing in voting responses due to missing `Expose()` decorator in dto.

### Change list

- Added Expose decorator to dto props


jkrizsan (approved):
Good Job!

PR #689, opened 2025-10-02
Buy item endpoint not adding errors to responses.

PlayJeri:
### Brief description

Possible `clanShop/buy` endpoint errors were missing from the response due to missing return statement in the method call.

### Change list

- Added return statement to return possible errors from buyItem call


jkrizsan (approved):
Looks Good

PR #693, opened 2025-10-04
Feature/add edit your avatar's appearance daily task 648

jkrizsan:
### Brief description

Add support for "edit your avatar's appearance" daily task.

### Change list

- Emit server event with CHANGE_AVATAR_OUTLOOK type if player.avatar objects changes
- Move event emitter code to a private method

### DIFF
--- a/src/player/player.controller.ts
+++ b/src/player/player.controller.ts
@@ -30,6 +30,7 @@ import { IncludeQuery } from '../common/decorator/param/IncludeQuery.decorator';
 import ApiResponseDescription from '../common/swagger/response/ApiResponseDescription';
 import EventEmitterService from '../common/service/EventEmitterService/EventEmitter.service';
 import { ServerTaskName } from '../dailyTasks/enum/serverTaskName.enum';
+import { isEqual } from 'lodash';
 
 @Controller('player')
 export default class PlayerController {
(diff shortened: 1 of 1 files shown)


PlayJeri (approved):
Good job.

jkrizsan:
@PlayJeri

I tried out this kind of simples solution unfortunately not working on my side (it was my original solution). If I am sending twice the same request it's detect the differences, regardless if t the request/DB are the same... Maybe because the DTO models are different and property order are different in the DTOs.

Is it working for you properly?

PlayJeri:
@jkrizsan Seems to be working fine for me. Try again and if it still fails give me a detailed explanation where and why it fails.

jkrizsan:
@PlayJeri tried again with the dev branch.

Initially, there is differences in one property request/DB

DB:
image

Request: differences in the head field

image

Event was emitted, what is correct:

image

DB was updated:

image

Sending the same request again:

image

And emit the event again, what I do NOT expect, the oldAvatar and newAvatar contains the head property with value 2

image

Root cause could be that the isEqual do a deep comparison, oldAvatar has AvatarDto type and the newAvatar has the ModifyAvatarDto type.

PlayJeri:
@jkrizsan You are correct. What an weird thing. When logged the player.avatar was a plain object and the body.avatar was ModifyAvatarDto {}.

So your solution of stringify + parse was actually the best option here.

javascript doing javascript things 😅

PR #694, opened 2025-10-04
Feature/add add an item to the flea market daily task 647

jkrizsan:
### Brief description

Add support for "add an item to the flea market" daily task.

### Change list

- Emit server event with ADD_ITEM_TO_FLEA_MARKET type when moving approved for sale item to the flea market
- Fix the failed tests


PlayJeri:
Other solutions are fine aswell but I think the possible multiple calls to eventemitter should come from same place

PR #698, opened 2025-10-05
Add new methods to box creator

PlayJeri:
### Brief description

Added two new methods to box creator to support the new way of creating them with teacher profile.

### Change list

- Add v2createBox method
- Add canCreateBox method
- Change box service basic service private -> public

### DIFF
--- a/src/box/boxCreator.ts
+++ b/src/box/boxCreator.ts
@@ -154,8 +154,11 @@ export default class BoxCreator {
    * @param boxName
    */
   public async v2createBox(teacherProfile_id: string, boxName: string) {
-    const [, maxLimitError] = await this.canCreateBox(teacherProfile_id);
-    if (maxLimitError) return [null, maxLimitError];
+    const [, validationErrors] = await this.v2validateBox(
+      teacherProfile_id,
+      boxName,
+    );
+    if (validationErrors) return [null, validationErrors];
 
     const boxToCreate_id = new ObjectId();
     const boxToCreate: Partial<v2Box> = {};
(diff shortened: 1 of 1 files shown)


src/box/box.service.ts
```diff
@@ -40,7 +40,7 @@ export class BoxService {
   }
 
   public readonly refsInModel: BoxReference[];
-  private readonly basicService: BasicService;
+  public readonly basicService: BasicService;
```
jkrizsan, line 43: I think this is better with private

src/box/boxCreator.ts
```diff
@@ -203,6 +239,30 @@ export default class BoxCreator {
     return [true, null];
   }
 
+  /**
+   * Checks if a box can be created
+   * @param teacherProfile_id
+   *
+   * @returns true if box can be created or ServiceErrors if found
+   */
+  private async canCreateBox(
```
jkrizsan, line 248: I think could be better to move this method to the boxHelper as a  public method. Rename it as validateBox2 (add more suitable name)

src/box/boxCreator.ts
```diff
@@ -142,6 +145,39 @@ export default class BoxCreator {
...
+      generateRandomClanName(),
+    );
+    const clanName2 = await this.uniqueFieldGenerator.generateUniqueFieldValue(
+      this.clanModel,
+      'name',
+      generateRandomClanName(),
+    );
+    boxToCreate.clansToCreate = [{ name: clanName1 }, { name: clanName2 }];
+
+    return await this.boxService.basicService.createOne(boxToCreate);
```
jkrizsan, line 178: So this was the reason why you made the BoxService.baseService public method. I think better to create a second BoxService.Create() method (which one use the canCreateBox method logic from the boxHelper for validation) and use here that new BoxService.Create()  method.

jkrizsan (commented):
Overall good Job. I added a couple of comments, how I think its should it refactor in order to be the code more reusable.

PR #702, opened 2025-10-07
addBoxIdToSchema changes

PlayJeri:
### Brief description

The plugin was adding the box_id to all schemas which was unnecessary. The teacher profile and box do not need that and neither do the subschemas like clansToCreate.


jkrizsan (approved):
Good Job

PR #705, opened 2025-10-07
Add v2 of configureBox endpoint

PlayJeri:
### Brief description

Added new version of the configureBox endpoint.

### Change list

- Create new endpoint
- Add is positive int validation to dto testersAmount


jkrizsan (approved):
Good Job.

Just fixed the imports for the BoxController. Also wanted to add @SwaggerTags('Box'), but this is already in use for other controller. I am not sure if the swagger tag should be unique or not. Can consider later on as well.

PR #712, opened 2025-10-19
Change leaderboard update interval

PlayJeri:
### Brief description

Leaderboard should update more often while the game is still in testing. I changed the static value for leaderboard update interval in leaderboard service to come from env vars so it's easier to change in the future depending the current needs. The current default value is now 1 minute.


jkrizsan (approved):
Good Job

PR #731, opened 2025-11-20
Change/remove get custom characters pagination

PlayJeri:
### Brief description

This PR removes the pagination from `CustomCharacterController` get all endpoint. Pagination is not required since the custom character amount will stay relatively small and pagination causes issues in the client side.

### Change list

- Removed the pagination
- Removed unused imports


jkrizsan (approved):
Looks Good.

What is the issue on the client side? Could be better to keep it.

PlayJeri:
> Looks Good.
> 
> 
> 
> What is the issue on the client side? Could be better to keep it.

The number of custom characters will stay below 100. So pagination isn't necessary. The issue is that client might have to make two requests instead of just one. So it's more convenient for them and they requested the change

PR #732, opened 2025-11-20
Update mongo image in init-script files to pull exact and correct version

hoan301298:
### Brief description

It caught error when firstly run command to set up the DB with pulling "mongo:8". The error was featureCompatibilityVersion - not support mongo usage version of 8.0.12-rc0-noble. Fix the pulling version as mongo:8.0.12-rc8-noble will make them match.

### Change list

- init-db-on-linux.sh
- init-db-on-windows.ps1


PlayJeri (approved):
Good work!

PR #736, opened 2025-12-02
Bug/reactions to global chat messages get returned as clan messages

PlayJeri:
### Brief description

This PR fixes the hard coded clan chat type to all chat message reactions. The chat type is now set dynamically from the db document.

### Change list

- Updated the `type` property in `ChatMessageDto` to use the `ChatType` enum instead of a string, enforcing stricter typing for chat message types.
- Modified `baseChat.service.ts` so that the `chat` field in the `ChatEnvelopeDto` is set to `updatedMessage.type` (which is now a `ChatType`), rather than being hardcoded to `ChatType.CLAN`, ensuring the correct chat type is used dynamically.
- Fix related test


jkrizsan (approved):
Looks good

PR #738, opened 2025-12-11
Validate message content no longer than 64 characters

hoan301298:
### Brief description

This PR validated message content's length in chatModule, the message content must not exceed 64 characters (Using prewritten messages to select).

### Change list

- update `createMessage.dto.ts` to add decorator `@MaxLength` to validate the length.

### DIFF
--- a/src/chat/dto/createMessage.dto.ts
+++ b/src/chat/dto/createMessage.dto.ts
@@ -32,14 +32,14 @@ export class CreateChatMessageDto {
   sender_id: string | ObjectId;
 
   /**
-   * Text content of the message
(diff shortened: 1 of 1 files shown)


jkrizsan (approved):
Looks good

PlayJeri (approved):
Nice work. Did you try the default error message? The library should give a similar default error message. It is ok to add your own but it might not be necessary

hoan301298:
> Nice work. Did you try the default error message? The library should give a similar default error message. It is ok to add your own but it might not be necessary

I was trying to find the default error message by testing the api.

Issue #739, opened 2025-12-12
Add tests for the friendship module.


PlayJeri:
### General description

The new friendship module needs tests. This includes the testing module, DTO builders and the unit tests / integration tests. The goal is to ensure that the new service works as expected.

### Things to notice

- Test should at least validate that db queries work as expected, errors are handled correctly and possible edge cases (there is no need to test the controller so the request and responses can be exluded).
- Injected services like the FriendshipNotifier can be mocked if needed.

### Tasks

- [ ] Create testing module for friendship service
- [ ] Create data builders for friendship related DTOs and documents
- [ ] Write tests for the friendship service

PR #748, opened 2025-12-17
Feature/add-tests-friendship-module-#739

hoan301298:
### Brief description
This PR implements all the tests for friendship module. Including module, DTO builders and unit tests for the friendship service and notifier to ensure the service work as expected. Also some changes of types, validation and handling error service in `friendship` schema, service. Following ticket #739 

### Change list

- Added `src/__test__/friendship/` folder including `modules`, `data` (builders) and `friendshipService`.
- `src/friendship/friendship.schema.ts`: Replace the type of `Mongoose.Types.ObjectId` to `string | ObjectId`. This change will make it being possible to create new runtime instance. Add validation for `requester` when status is `PENDING`.
- `src/friendship/friendship.service.ts`: Add return type for `getFriendRequests()` and return `[null, error]` when the `NotFoundServiceError` has been returned from `getFriendshipsWithStatus()` instead of throw the error.

### DIFF
--- a/src/__tests__/friendship/data/builder/FriendRequestDtoBuilder.ts
+++ b/src/__tests__/friendship/data/builder/FriendRequestDtoBuilder.ts
@@ -1,0 +1,32 @@
+import { ObjectId } from "mongodb";
+import IDataBuilder from "src/__tests__/test_utils/interface/IDataBuilder";
+import { FriendRequestDto } from "src/friendship/dto/FriendRequest.dto";
+
+export default class FriendRequestDtoBuilder
+    implements IDataBuilder<FriendRequestDto>
+{
+    private readonly base: FriendRequestDto = {
+        friendship_id: new ObjectId().toString(),
+        direction: 'incoming', // 'outgoing'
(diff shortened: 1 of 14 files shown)


jkrizsan (commented):
Do you want to merge it into the main branch?

hoan301298:
> Do you want to merge it into the main branch?

I changed it to dev branch, thanks for the notice!!

PlayJeri (approved):
Nice work. 

I see that you added the `model.deleteMany` call to beforeEach in tests. There is the global beforeEach (src/__tests__/test_utils/before-and-after-tests.ts) that should reset the database after each test. So it might not be necessary to add them to all test files.

hoan301298:
> Nice work.
> 
> I see that you added the `model.deleteMany` call to beforeEach in tests. There is the global beforeEach (src/**tests**/test_utils/before-and-after-tests.ts) that should reset the database after each test. So it might not be necessary to add them to all test files.

Thanks for this advice Jeri.

I just cleaned them.

PR #749, opened 2025-12-18
Feature/add transaction box

jkrizsan:
### Brief description

Lets try to add the transaction support to the modules one by one. This PR created for the Box module.
Main purpose of this PR is to validate the concept!

### Change list

- Add Transactions.ts with all of the required methods.
- Apply the Transactions method for the classes where it was possible.


PlayJeri (commented):
I like the idea of the transaction helper methods but I think there is some need for improvement in them and we need core rules on how to handle the transactions in the project. So this is what I think.

1. We should drop the openedSession parameter. openedSession exists to decide at runtime whether a helper "owns" the session or not. This makes transaction ownership implicit and conditional, which might cause problems. I think we need a rule that if a function creates a session, it owns the transaction lifecycle and if a function receives a session, it must never commit, abort or end it. Once this rule is enforced the opened session is no longer needed and the transaction flow is simpler to understand.
2. Add try / catch /finally to transaction helpers. MongoDB session methods can fail. So `commitTransaction`, `abortTransaction` and `endSession` can throw. Without catch and finally blocks there can be silent failures or memory leaks due to sessions staying open. I think we should also start using loggers in these types of functions so we could debug the possible errors in production.
3. Make helper methods transaction-agnostic. So helpers methods should never start or end transaction but only accept a `ClientSession` and operate within it.

``` ts
export async function cancelTransaction(
  session: ClientSession,
  errors: ServiceError | ServiceError[],
): Promise<IServiceReturn<never>> {
  const errorsArray = Array.isArray(errors) ? errors : [errors];

  try {
    if (session.inTransaction()) {
      await session.abortTransaction();
    }
  } catch (abortError) {
    logger.error('Failed to abort transaction', abortError);
  } finally {
    try {
      await session.endSession();
    } catch (endError) {
      logger.error('Failed to end session', endError);
    }
  }

  return [null, errorsArray];
}
```

```ts
export async function endTransaction<T = true>(
  session: ClientSession,
  value?: T,
): Promise<IServiceReturn<T | true>> {
  const result =
    typeof value === 'undefined' ? (true as const) : value;

  let commitError: unknown;

  try {
  } catch (err) {
    commitError = err;
  } finally {
    try {
      await session.endSession();
    } catch (endError) {
      logger.error('Failed to end session', endError);
    }
  }

  if (commitError) {
    logger.error('Commit failed', commitError);
    return [null, [new TransactionCommitError(commitError)]];
  }

  return [result, null];
}
```
Here are some examples. So I'd like to hear your thoughts on this.

hoan301298:
- I agree with all of those points. Making transaction ownership explicit really improves clarity and reduces the risk of subtle bugs around commits, aborts, or ending sessions at the wrong time.

- Adding `try` / `catch` / `finally` around transaction helpers is a great point to prevent issues like unhandled errors or sessions not being properly closed. Using loggers feels like a practical improvement that increases reliability without adding much complexity, so I think these changes should be included.

PR #750, opened 2026-01-07
Feature/add-transaction-box

### LINKED ISSUE #742: Add transaction support to Box module
### Links

[Swagger]()

[Flow chart]()

### Tasks

- [ ]

hoan301298:
### Brief description

This PR adds transaction supports for Box module and apply the changes for some classes using transaction methods.
Issue: #742

### Change list

- Drop `openedSession`: prevents problems. Basically, a function creates a session then it owns the transaction lifecycle.
- Add `try`/`catch`/`finally`: catches errors when working with MongoDB and return them easily.
- Apply changes to the use of transaction methods in classes (`Box` module)
- Add `TransactionCommitError.ts`, `Transaction.ts`.
- ESLint


src/common/function/Transactions.ts
```diff
@@ -0,0 +1,100 @@
...
+import TransactionCommitError from './TransactionCommitError';
+
+const logger = new Logger();
+
+/**
+ * Initializes and starts a database session for transactions.
+ *
+ * @param connection - Mongoose database connection.
+ *
+ * @param openedSession - (Optional) An already opened ClientSession to use.
```
PlayJeri: Not needed anymore

src/common/function/Transactions.ts
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @returns A promise that resolves to the started database session.
+ */
+export async function initializeSession(
+  connection: Connection,
+  openedSession?: ClientSession,
```
PlayJeri: The openedSession param can be dropped since we agreed that the parent method that initializes the transaction must be the one to close it.

src/common/function/Transactions.ts
```diff
@@ -0,0 +1,100 @@
...
+ * Aborts the database transaction and ends the session.
+ *
+ * @param session - Started database session.
+ * @param errors - The error to be returned.
+ * @returns A promise that returns an unexpected service error.
+ */
+export async function cancelTransaction(
+  session: ClientSession,
+  errors: ServiceError | ServiceError[],
+): Promise<IServiceReturn<never>> {
```
PlayJeri: I think that the type should just be void. So `Promise<IServiceReturn<void>>` or `Promise<void>` if the errors param gets dropped.

src/common/function/Transactions.ts
```diff
@@ -0,0 +1,100 @@
...
+    logger.error('Failed to abort transaction', abortError);
+  } finally {
+    try {
+      await session.endSession();
+    } catch (endError) {
+      logger.error('Failed to end transaction', endError);
+    }
+  }
+
+  return [null as never, errorsArray];
```
PlayJeri: So here the return can be [null, errorsArray];

src/common/function/Transactions.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/**
+ * Aborts the database transaction and ends the session.
+ *
+ * @param session - Started database session.
+ * @param errors - The error to be returned.
+ * @returns A promise that returns an unexpected service error.
+ */
+export async function cancelTransaction(
```
PlayJeri: You can also consider is the errors param even needed here. Since the method doesn't actually do anything with that param. It gets returned exactly the same as it entered no matter what happens inside this method.

src/common/function/Transactions.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/**
+ * Commits the database transaction and ends the session.
+ *
+ * @param session - Started database session.
+ * @param returnValue - (Optional) Value to return upon successful transaction completion.
+ * @returns A promise that resolves to a successful service return.
+ */
+export async function endTransaction(
```
PlayJeri: I think the function overloading here is unnecessary you could just keep the last one the `endTransaction<T = true>(session: ClientSession, value?: T)

src/box/accountClaimer/accountClaimer.service.ts
```diff
@@ -31,13 +37,16 @@ export default class AccountClaimerService {
...
+   * @param openedSession - (Optional) An already opened ClientSession to use.
+   *
    * @returns Claimed account data, as well as an access token, or ServiceErrors:
    * - REQUIRED - if the password is not provided
    * - NOT_FOUND - if there are no box with this password
    * - NOT_AUTHORIZED - if there are no places left (testersAmount < testersAccountsClaimed)
    */
   async claimAccount(
     password: string,
+    openedSession?: ClientSession,
```
PlayJeri: Helper methods shouldn't have optional session parameters.

PlayJeri (changes requested):
So I think the transaction helper methods can be simplified a little bit. I added more specific comments about them. Then the bigger idea behind this whole transaction support change was to implement a better Transaction ownership principle. Meaning that the method that starts / initializes the transaction must be the one that cancels or commits it. So the helper methods that take in the session shouldn't call the cancelTransaction. They only take the session as param so that the session can be passed to the `BasicService` so that the db operations are all under the same transaction. But the parent method is responsible for that transaction lifecycle.

And maybe consider matching naming for the transaction helpers. So initialize-, abort-, commitTransaction or Session. A bit nitpicky but might be easier to follow if all helpers are called somethingSession or somethingTransaction.

src/common/function/Transactions.ts
hoan301298: I believe that passing the errors here might be a good way. Examples in below that they need to return the error right away when the error is occurred, then definitely our code would be easy to return it. So in my opinion, we should keep it to make our function cleaner.

```typescript
    if (updateErrors && updateErrors[0].reason === SEReason.NOT_FOUND)
    return  await cancelTransaction(session);

    if (updateErrors)
    return  await cancelTransaction(session);

    if (readErrors) 
    return  await cancelTransaction(session);
```

src/common/function/Transactions.ts
hoan301298: I have added ```try-catch``` in case ```startSession()``` and ```startTransaction()``` can fail. How do you think?

```typescript
export async function initializeSession(
  connection: Connection,
): Promise<IServiceReturn<ClientSession>> {
  let session: ClientSession | null = null;

  try {
    session = await connection.startSession();
    session.startTransaction();
    return [session, null];
  } catch (startError) {
    logger.error('Failed to start Mongo transaction', startError);
    if (session) await session.endSession();

    return [
      null,
      [
        new ServiceError({
          reason: SEReason.UNEXPECTED,
          field: 'session',
          value: startError,
          message: 'Failed to start Mongo transaction',
        }),
      ],
    ];
  }
}
```

src/common/function/Transactions.ts
hoan301298: I mentioned it in below.

PlayJeri (approved):
Well done. Good work

Issue #751, opened 2026-01-10
Fix ESM compatibility for 'webdav' package in CommonJS context


CapoMK25:
## `[8:35:17 PM] Starting compilation in watch mode...

[8:35:26 PM] Found 0 errors. Watching for file changes.

[dotenv@17.2.2] injecting env (0) from .env -- tip: ⚙️  specify custom .env file path with { path: '/custom/path/.env' }
C:\Users\Matias.DESKTOP-2SBTD6B\Desktop\Altzone-Server\dist\gameAnalytics\logFile.service.js:19
const webdav_1 = require("webdav");
                 ^

Error [ERR_REQUIRE_ESM]: require() of ES Module C:\Users\Matias.DESKTOP-2SBTD6B\Desktop\Altzone-Server\node_modules\webdav\dist\node\index.js from C:\Users\Matias.DESKTOP-2SBTD6B\Desktop\Altzone-Server\dist\gameAnalytics\logFile.service.js not supported.
Instead change the require of index.js in C:\Users\Matias.DESKTOP-2SBTD6B\Desktop\Altzone-Server\dist\gameAnalytics\logFile.service.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (C:\Users\Matias.DESKTOP-2SBTD6B\Desktop\Altzone-Server\dist\gameAnalytics\logFile.service.js:19:18) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v20.11.0

This is the error output. The fix has already been done by me. This issue is just for documentation purposes.

CapoMK25:
Discovered that updating npm/node from the antique version 20 fixes this issue completely. No changes needed here, closing this issue.

PR #753, opened 2026-01-10
Fix(WebDav): add dynamic import helper to support ESM in CJS context …

CapoMK25:
Brief description

When running `npm run start:dev` I ran into this issue:

```
`[8:35:17 PM] Starting compilation in watch mode...

[8:35:26 PM] Found 0 errors. Watching for file changes.

[dotenv@17.2.2] injecting env (0) from .env -- tip: ⚙️  specify custom .env file path with { path: '/custom/path/.env' }
C:\Users\Matias.DESKTOP-2SBTD6B\Desktop\Altzone-Server\dist\gameAnalytics\logFile.service.js:19
const webdav_1 = require("webdav");
                 ^

Error [ERR_REQUIRE_ESM]: require() of ES Module C:\Users\Matias.DESKTOP-2SBTD6B\Desktop\Altzone-Server\node_modules\webdav\dist\node\index.js from C:\Users\Matias.DESKTOP-2SBTD6B\Desktop\Altzone-Server\dist\gameAnalytics\logFile.service.js not supported.
Instead change the require of index.js in C:\Users\Matias.DESKTOP-2SBTD6B\Desktop\Altzone-Server\dist\gameAnalytics\logFile.service.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (C:\Users\Matias.DESKTOP-2SBTD6B\Desktop\Altzone-Server\dist\gameAnalytics\logFile.service.js:19:18) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v20.11.0
```

Turns out the project doesn't work out of the gate and required extensive tweaking to the source code to bypass this ESM vs CommonJS mismatch. This fixes said compatibility issue so that the issue doesn't repeat for anybody else. Otherwise this is a critical blocker for anyone just cloning the repo and following the README instructions.

Change list
- Added a helper utility to dynamically import ESM modules (createWebDavClient.ts)
- Updated logFile.service.ts to use the new helper
- TSConfig was adjusted (module: es2020) to allow import.meta.url usage safely

## Why this matters
Without this fix, the app throws ERR_REQUIRE_ESM when importing `webdav` using `require()` in compiled output as we can see in the output.

- Closes https://github.com/Alt-Org/Altzone-Server/issues/751

### DIFF
--- a/src/common/utils/createWebDavClient.ts
+++ b/src/common/utils/createWebDavClient.ts
@@ -1,0 +1,14 @@
+// ESM-only workaround: webdav is ESM-only, so we use dynamic import to avoid CommonJS issues
+import { createClient, WebDAVClient } from 'webdav';
+import { envVars } from '../service/envHandler/envVars';
+
+export function createWebDavClient(): WebDAVClient {
+  return createClient(
+    `http://${envVars.OWNCLOUD_HOST}:${envVars.OWNCLOUD_PORT}/remote.php/webdav/`,
+    {
+      username: envVars.OWNCLOUD_USER,
+      password: envVars.OWNCLOUD_PASSWORD,
+      maxBodyLength: 52428800,
+    },
+  );
+}
--- a/src/profile/dto/playerProfile.dto.ts
+++ b/src/profile/dto/playerProfile.dto.ts
@@ -13,5 +13,6 @@ export class PlayerProfileDto extends CreatePlayerDto {
   @IsProfileExists()
   @IsMongoId()
   @IsOptional()
-  override profile_id: string;
+  override profile_id = '';
+
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
         "mongoose": "8.19.4",
         "mqtt": "5.14.1",
         "rxjs": "7.8.2",
-        "webdav": "5.8.0",
+        "webdav": "^4.11.2",
         "ws": "8.18.3"
       },
       "devDependencies": {
@@ -735,15 +735,6 @@
       "funding": {
         "type": "github",
         "url": "https://github.com/sponsors/Borewit"
-      }
-    },
-    "node_modules/@buttercup/fetch": {
-      "version": "0.2.1",
-      "resolved": "https://registry.npmjs.org/@buttercup/fetch/-/fetch-0.2.1.tgz",
-      "integrity": "sha512-sCgECOx8wiqY8NN1xN22BqqKzXYIG2AicNLlakOAI4f0WgyLVUbAigMf8CZhBtJxdudTcB1gD5lciqi44jwJvg==",
-      "license": "MIT",
-      "optionalDependencies": {
-        "node-fetch": "^3.3.0"
       }
     },
     "node_modules/@casl/ability": {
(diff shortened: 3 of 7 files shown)


src/profile/dto/playerProfile.dto.ts
PlayJeri, line 1: What is the purpose of this change?

Overriding profile_id as empty string will cause the `@IsMongoId` and `@IsProfileExists` validation to fail.

src/common/utils/createWebDavClient.ts
PlayJeri, line 1: This helper is not used anywhere?

PlayJeri (changes requested):
I found this PR quite confusing and after finding some issues I Run a AI review on it and it gave the following recommendations.

Summary
The PR intent says it fixes runtime ERR_REQUIRE_ESM for webdav, but the current code still uses static imports from webdav, so the error will persist if an ESM-only version is installed. There is also inconsistency between code and package.json, and the helper is not correctly used.

Below are the required fixes to make this consistent and actually resolve the ESM issue.

✅ Required fixes
1. Use ESM version of webdav

package.json

"webdav": "^5.8.0"

2. Fix helper to use dynamic import (no static imports)

src/common/utils/createWebDavClient.ts

import { envVars } from '../config/env';

export async function createWebDavClient() {
  const { createClient } = await import('webdav');

  const baseUrl = `http://${envVars.OWNCLOUD_HOST}:${envVars.OWNCLOUD_PORT}/remote.php/webdav/`;

  return createClient(baseUrl, {
    username: envVars.OWNCLOUD_USER,
    password: envVars.OWNCLOUD_PASSWORD,
    maxBodyLength: 52_428_800,
  });
}


❗ Do not statically import anything from 'webdav' (including type imports), otherwise Node may still attempt require() at runtime.

3. LogFileService: remove all webdav imports and use helper

Remove

import { createClient, WebDAVClient } from 'webdav';


Add

import { createWebDavClient } from '../common/utils/createWebDavClient';


onModuleInit

async onModuleInit(): Promise<void> {
  this.client = await createWebDavClient();
}

4. Upload using Buffer (no stream needed)

Simpler and supported in webdav v5:

await this.client.putFileContents(
  filePath,
  fileToSave.buffer,
  { overwrite: true },
);


Remove bufferToStream / Readable usage unless there is a strong reason to stream.

⚠️ Additional bug (unrelated but should be fixed)
PlayerProfileDto validation
profile_id = '';


conflicts with:

@IsOptional()
@IsMongoId()


Empty string is not optional and will fail validation.

Fix:

profile_id?: string;


(or remove the default value entirely)

ℹ️ Notes

Dynamic import() works in CommonJS builds; no need to convert the whole project to ESM.

Current PR mixes ESM workaround language with non-ESM code paths, which is why the runtime error will still occur.

PlayJeri:
The issue is most likely with you Node version. I tried to run the project with the v20.11.0 and I run into the same error. So update your node to latest 24 version and try again. @CapoMK25

hoan301298:
My setup worked fine with Node.js v22.13.0 and the same issue was caught when I changed to v20..
Let's try to install the latest node version on your local machine with:

```powershell
# install Node.js v24.13.0
nvm install 24
nvm use 24
```

PR #767, opened 2026-01-30
760 Feature request update the Player Avatar data to include color data for each of the pieces

CapoMK25:
### Brief description

Avatar Migration to an Object-Based Structure

This PR solves the feature request to add the Player Avatar data to include color data for each of the pieces.

### Change list

- Shifted avatar parts from `number` to `{ id, color }` objects.
- Updated `ModifyAvatarDto` and `UpdatePlayerDto` to handle nested validation.
- Adjusted the clothes-change detection logic to compare `.id` properties.
- Updated `createPlayerDtoBuilder` and related mocks to fix the 32 failing suites caused by the type mismatch.


CapoMK25:
JSDocs documentation went missing along the way for some reason again... added it back.

hoan301298 (approved):
Looks good to me. To be clearer, I think having IDs here should have some kind of meaning, or it's simply an arrangement from 1 to 9 for each properties inside.

src/__tests__/player/data/player/createPlayerDtoBuilder.ts
```diff
@@ -16,15 +16,15 @@ export default class CreatePlayerDtoBuilder
...
-      hands: 1,
+      head: { id: 1, color: '#ffffff' },
+      hair: { id: 1, color: '#ffffff' },
+      eyes: { id: 1, color: '#ffffff' },
+      nose: { id: 1, color: '#ffffff' },
+      mouth: { id: 1, color: '#ffffff' },
+      eyebrows: { id: 1, color: '#ffffff' },
+      clothes: { id: 1, color: '#ffffff' },
+      feet: { id: 1, color: '#ffffff' },
+      hands: { id: 1, color: '#ffffff' },
```
hoan301298, line 27: Based on what we have here, I don't see how the IDs convey any particular meaning. This is something we should clarify with the UI team, and whether IDs really need to be unique.

PR #772, opened 2026-02-05
Automatic clan join (ad-hoc request from the UI Team)

CapoMK25:
### Brief description

The UI-Team asked the server team on discord to explain that the automatic clan join when creating an account doesn't work if all clans are full. They asked for a new clan to be added (or multiple to be exact).

So this PR implements an automatic fallback to ensure every new player is onboarded into a clan immediately. Previously, new players would've remained clanless if all existing clans reached the 30-player limit.

### Change list

- Logic Implementation (`JoinService`)
- Capacity Fallback: Added a secondary check within the `player.created` event listener. If no isOpen clans with `room (playerCount < 30)` are found, the system now triggers an automated clan creation.

- Smart Naming: Implemented a dynamic naming convention: `Expedition [TotalClans + 1]-[HHmm]`. This prevents naming collisions and provides a sense of progression.

- Infrastructure Reuse: Utilized `createOneWithoutAdmin` to ensure all standard clan side-effects (SoulHome initialization, default Roles, and Stock) are executed correctly.

- DTO Compliance: Refactored the creation logic to use `CreateClanDto`, satisfying mandatory fields like phrase and labels without using as any type-casting.

- Error Handling: Wrapped the creation process in a `try-catch` block with NestJS Logger integration. This ensures that if the auto-creation fails, it won't crash the player-creation lifecycle, while still leaving a trace for debugging later due to the logger.

- Spec Update: Updated `findClanForNewPlayer.test.ts`. The test formerly asserting that a player remains clanless in a full world now asserts that an "AUTO" tagged clan is created and joined.

- Regression Testing: Verified that all 1,153 tests pass, ensuring no regressions in the standard join or voting flows.

- Manual Test: Verified on the local environment by setting all clans to `playerCount: 30` manually.

- Result: A new player was created, an Expedition clan was generated, and the player was successfully linked via `clan_id`.


src/clan/join/join.service.ts
```diff
@@ -186,15 +188,49 @@ export class JoinService {
    */
   @OnEvent('player.created')
   async findClanForNewPlayer(playerId: string) {
-    const randomClan = await this.clanService.model
```
hoan301298, line 189: This function finds a random `clan` matching criteria.

src/clan/join/join.service.ts
```diff
@@ -186,15 +188,49 @@ export class JoinService {
...
-      .aggregate<ClanDocument>([
-        { $match: { isOpen: true, playerCount: { $lt: 30 } } },
-        { $sample: { size: 1 } },
-      ])
-      .then((res) => res[0]);
 
-    if (!randomClan) return;
+  const availableClan = await this.clanService.model.findOne({
```
hoan301298: But your solution finds the first document by index order --> so technically, if 2 players are created and they will be all added to a `clan` including <29 players. It's better to add them randomly, so we all have some newbie distributed evenly across different teams.

src/clan/join/join.service.ts
```diff
@@ -186,15 +188,49 @@ export class JoinService {
...
 
-    await this.joinClan(playerId, randomClan._id.toString());
+  if (availableClan) {
+    return await this.joinClan(playerId, availableClan._id.toString());
   }
+
+  const joinableClanCount = await this.clanService.model.countDocuments({
+    isOpen: true,
+    playerCount: { $lt: 30 }
+  });
```
hoan301298: This is so much extra work, so the logic is `if (availableClan)` existed, let player join into it, else - create a new `clan`. Therefore, this function doesn't help and check block`if (joinableClanCount > 0)` should be removed too

src/clan/join/join.service.ts
```diff
@@ -186,15 +188,49 @@ export class JoinService {
...
+
+  if (joinableClanCount > 0) {
+    return;
+  }
+
+  try {
+    const totalClans = await this.clanService.model.countDocuments();
+    const now = new Date();
+    const suffix = `${now.getHours()}${now.getMinutes()}`; 
+    const newClanName = `Expedition ${totalClans + 1}-${suffix}`;
```
hoan301298: I think it's better to ask UI Team for the rule of naming random new clan - for example: "New Clan-1", "New Clan-2". Try to avoid collision.

src/clan/join/join.service.ts
```diff
@@ -186,15 +188,49 @@ export class JoinService {
...
+
+    const [newClan, errors] = await this.clanService.createOneWithoutAdmin(createClanDto);
+
+    if (errors || !newClan) {
+      return;
+    }
+
+    await this.joinClan(playerId, newClan._id.toString());
+    
+  } catch (err: any) {
```
hoan301298: seems like `try-catch` block is unecessary here.

hoan301298 (changes requested):
I think your solution is quite good, only some extra-work might not be needed.

hoan301298 (approved):
Well done!

PR #774, opened 2026-02-09
766 Issue enforce game version compatibility to prevent matchmaking across different versions

### LINKED ISSUE #766: Issue: Enforce game version compatibility to prevent matchmaking across different versions
Background

During development and testing, multiple client versions may be active at the same time (e.g. older builds, teaching demos, open demos, internal test builds).
Currently, there is a risk that players running different game versions could end up in the same matchmaking pool or match, leading to desyncs, invalid results, or crashes.

Goal

Ensure that players running different game versions cannot match with each other, either directly or via matchmaking.

Requirements (server-side)
1. Client version reporting

The client must report its game version to the server.
This should be included in at least:

authentication / session creation, or

matchmaking / battle start registration

Example fields:

clientVersion (string, e.g. "0.6.2")

optionally: buildNumber or versionHash (for stricter matching)

2. Version-scoped matchmaking

Matchmaking must be strictly scoped by version:

Players are only matched with others using the same compatible version

Players with incompatible versions must never:

enter the same match

see each other as available opponents

share matchmaking queues

Recommended rule (MVP):

Exact version match only (clientVersion === clientVersion)

Possible future extension:

Compatibility ranges (e.g. 0.6.x), but not required now

3. Battle registration validation

When registering a match (e.g. /gamedata/battle/start):

All players in the match must have the same validated client version

If versions differ:

reject the registration

return a clear error code (e.g. VERSION_MISMATCH)

4. Result submission validation

When receiving battle results:

The submitted result must match the version associated with the match

Results from a client with a different version than the registered match version must be rejected

5. Error handling (client-facing)

If a version mismatch occurs:

Server should return a clear error response

Client can then:

prompt the user to update

or block matchmaking gracefully

Acceptance Criteria

Players on different game versions never enter the same match.

Matchmaking queues are isolated by game version.

Battle start fails if player versions do not match.

Battle results are rejected if the submitting client version does not match the match version.

Version mismatch errors are explicit and debuggable.

Implementation notes (non-binding)

Store clientVersion (or buildId) as part of:

session

matchmaking queue entry

match metadata

The simplest and safest MVP approach is exact version matching.

Avoid relying on client-side checks only; validation must be server-enforced.

This task is primarily about data integrity and stability during parallel development and testing, and it should be treated as a hard server constraint, not a soft recommendation.

CapoMK25:
### Brief description

This PR introduces `client_version` tracking for online players. This field allows the system to identify the game build version during the "ping" process, ensuring better matchmaking isolation and helping to prevent desyncs between incompatible client versions.

This change was requested from the lead of the project to ensure that incompatible versions wouldn't be able to match with each other.
The data is persisted in the Redis cache and is available for filtering when retrieving the list of online players.

### Change list

- `Domain Model Update`: Added `client_version` to the `OnlinePlayer `domain class and applied `@Expose()` decorators to ensure property visibility during serialization.

`DTO Validation`: Updated the `AddOnlinePlayer` payload class with `@IsString()` and `@IsOptional()` decorators to validate incoming client version data from the /ping endpoint.

Service Logic:

Modified `addPlayerOnline` to extract `client_version` from the request and include it in the Redis payload.

Updated `getOnlinePlayers` to support optional filtering by `client_version`.

Infrastructure & Cache

Redis Payload: Updated the JSON schema stored in Redis to include the new version string.
output

Test & Verification

Manual Verification: Confirmed via server logs that `client_version` is correctly received by the service and retrieved from the cache.

Postman Testing: Verified that sending `client_version` in a POST request results in the data being successfully processed by the backend.


CapoMK25:
Jenkins choked yesterday for some odd reason and took ages to finish. I re-ran it today just in case and now the CI is all green and ready for review.

src/onlinePlayers/battleQueue/battleQueue.service.ts
```diff
@@ -62,6 +62,21 @@ export class BattleQueueService {
...
 
+    const firstVersion = validPlayers[0].client_version;
+    const hasVersionMismatch = validPlayers.some(p => p.client_version !== firstVersion);
+
+    if (hasVersionMismatch) {
+      return [
+        null,
+        [
+          new ServiceError({
+            reason: SEReason.VALIDATION,
```
hoan301298: I suggest that we should create a new type for SEReason with name 'VERSION_MISMATCH' instead of using `VALIDATION`

hoan301298 (changes requested):
I think with the general infomation provided on the issue #766 , we should get more clearer information from UI team to match the query or body when UI sends data to server with `client_version` attached.

src/onlinePlayers/battleQueue/battleQueue.service.ts
CapoMK25: Addressed in my latest commit, good point.

src/onlinePlayers/battleQueue/battleQueue.service.ts
```diff
@@ -72,7 +72,7 @@ export class BattleQueueService {
         null,
         [
           new ServiceError({
-            reason: SEReason.VALIDATION,
+            reason: SEReason.VERSION_MISMATCH,
```
tickBit, line 75: I looked the entire implementation of this commit. Looks logical 👍

tickBit (approved):
I far as I can tell the requested changes are implemented. Good work 👍 Approved.

hoan301298 (approved):
This PR is good to go 👍 These new changes should be tested by UI Team and maybe ask them to give us feedback.
One more important thing: the `client_version` needs to be detailed and updated on Swagger.

PR #775, opened 2026-02-11
637 Feature: Daily Player Emotions

### LINKED ISSUE #637: Feature request: Player data should contain information about the daily emotions for the past week.
POST: /player/emotion
Takes an emotion as an enum in the body.

Returns the emotions list together with the player data from GET /player.

Additionally, we could have a endpoint to check if the emotion has already been sent today.
GET: /player/emotioncheck

When the game is opened for the first time that day the game should ask what the emotion of the player for that day is. 
For this the emotioncheck endpoint could be used to check if player has already sent the emotion.
If the emotion is not yet set then the player can send their selected emotion to server.

This could use the typical PUT: /player but having a separate endpoint would likely be more elegant and the server side handling is likely easier to set up this way.
This would cause the sent emotion to be saved into the emotions list as the emotion for the current day. If the player doesn't send the emotion for a specific day then that day will be tagged as blank.

This list is then returned when the player call GET for their player data together with the other information.

## Any additional info?

Enums:
Sorrow
Anger
Joy
Playful
Love
Blank (when emotion is not set.)

CapoMK25:
### Brief description

The UI Team lead Niko opened an issue last year regarding adding emotions to the backend API. This has now been implemented, unit tested and tested on Postman as well on this PR. 

### Change list
- Added `PlayerEmotion` enum and updated `PlayerSchema` with a history array.

- Implemented `GET /player/emotioncheck` to let the client know if the daily popup is needed.

- Implemented `POST /player/emotion` to record daily sentiment (limited to once per day).

- Fix: Resolved path alias issues in tests by switching to relative imports in the `Schema.`

- Fix: Resolved controller routing collision by reordering static routes above dynamic parameters.

- Fix: Aligned controller with `req.user.player_id` to ensure correct database targeting.


src/player/player.controller.ts
```diff
@@ -64,16 +63,42 @@ export default class PlayerController {
...
 
+  /**
+   * Checks if the authenticated player has already submitted an emotion for the current day.
+   * This is used by the game client on startup to determine if the emotion selection
+   * popup should be displayed or not.
+   * @param req - The request object containing the authenticated user's data.
+   * @returns An object containing `sentToday` (boolean).
+   */
+  @Get('/emotioncheck')
+  async checkDailyEmotion(@Req() req) {
```
hoan301298: Here, try to see how controller uses some decorator like `@UniformResponse()` ,`@LoggedUsers` or `@Authorize()`.

Using `@Req()` is an old way and not following our set of rules.

Also `@ApiResponseDescription()` is always required for showing on Swagger the detail of the Endpoint (success, error).

Check the return type should match with `IServiceReturn<boolean>` for GET `/emotioncheck`.

src/player/schemas/player.schema.ts
```diff
@@ -5,9 +5,15 @@ import { ExtractField } from '../../common/decorator/response/ExtractField';
...
 import { Avatar, AvatarSchema } from './avatar.schema';
+import { PlayerEmotion } from '../../common/enum/playerEmotion.enum';
 
 export type PlayerDocument = HydratedDocument<Player>;
 
+export interface IPlayerEmotion {
+  emotion: PlayerEmotion;
+  date: Date;
+}
+
```
hoan301298: This should be placed somewhere else to keep the clean structure of schema

src/player/enum/playerEmotion.enum.ts
hoan301298, line 1: Why this file is in common folder?

src/player/schemas/player.schema.ts
```diff
@@ -64,6 +70,22 @@ export class Player {
...
+          enum: Object.values(PlayerEmotion),
+          required: true,
+        },
+        date: { type: Date, default: Date.now },
+      },
+    ],
+    _id: false,
+    default: [],
+  })
+  emotions?: { emotion: PlayerEmotion; date: Date }[];
```
hoan301298: Maybe you can think if we can set the default would be BLANK and that's easier for you to check it in workflow and logic.

src/player/player.controller.ts
```diff
@@ -64,16 +63,42 @@ export default class PlayerController {
...
+   * @param body - The DTO containing the selected PlayerEmotion enum value.
+   * @returns The updated player document including the new emotion entry.
+   * @throws BadRequestException - If the player attempts to submit more than once per day.
+   */
+  @Post('/emotion')
+  @UniformResponse(ModelName.PLAYER, PlayerDto)
+  async setDailyEmotion(@Req() req, @Body() body: UpdateEmotionDto) {
+    return this.service.addEmotion(req.user.player_id, body.emotion);
+  }
+
```
hoan301298, line 107: Same with GET issue. All of these endpoint need to have `authenticated` session.

src/player/player.service.ts
```diff
@@ -248,4 +249,48 @@ export class PlayerService
     );
     if (updateErrors) throw updateErrors;
   }
+
+  /**
+   * Checks if the player has already submitted an emotion today.
+   * @param playerId - The unique identifier of the player.
+   * @returns Boolean indicating if an entry for today exists.
+   */
+  async checkIfEmotionSentToday(playerId: string): Promise<boolean> {
```
hoan301298: Method should return as `IServiceReturn<boolean>` like success: `[true, null]` and if error: `[null, errors]`

hoan301298 (changes requested):
This PR needs to be fixed and should cover all of following rules.

src/player/schemas/player.schema.ts
CapoMK25: Do you mean the interface or something else?

hoan301298: Yes, you can create a new file to put the interface in, the schema should be isolated from other parts, even though it owns.

This will keep our schema looks clean.

src/player/player.service.ts
CapoMK25: Addressed in my commit just now.

src/player/player.controller.ts
CapoMK25: Addressed in my commit just now.

src/player/schemas/player.schema.ts
CapoMK25: Addressed in my commit just now. Good point, I agreed.

src/player/enum/playerEmotion.enum.ts
CapoMK25: Addressed in my commit just now.

src/player/schemas/player.schema.ts
CapoMK25: Addressed in my commit just now.

src/player/player.controller.ts
CapoMK25: Addressed in my commit just now.

src/player/player.service.ts
```diff
@@ -248,4 +253,73 @@ export class PlayerService
...
   }
-}
+
+  /**
+   * Checks if the player has already submitted an emotion today.
+   * @param playerId - The unique identifier of the player.
+   * @returns - A classic tuple setup [boolean, ServiceError[]] indicating if an entry for today exists.
+   */
+  async checkIfEmotionSentToday(playerId: string): Promise<IServiceReturn<EmotionCheckDto>> {
+    const player = await this.model
```
hoan301298: I suggest, it's better if you can use `BasicService` functions.

So maybe this function should be return `<IServiceReturn<boolean>>`.

src/player/player.controller.ts
```diff
@@ -64,16 +66,53 @@ export default class PlayerController {
...
+  * Registers the player's selected emotion for the current day.
+  */
+  @ApiResponseDescription({
+  success: { dto: PlayerDto, modelName: ModelName.PLAYER, status: 201 },
+  errors: [400, 401, 403, 409],
+  })
+  @Post('/emotion')
+  @UniformResponse(ModelName.PLAYER, PlayerDto)
+  @Authorize({ action: Action.create, subject: PlayerDto })
+  public async setDailyEmotion(
```
hoan301298, line 96: This controller function should return nothing, since you return 201 status_code (means `No Content`).

hoan301298 (approved):
There are a few issues needed to fix in this PR. But in overall, it looks good to me. Approved!

CapoMK25:
@hoan301298 or @Iir0R you can check and approve once more before I hit merge on this, thanks!

src/player/player.service.ts
CapoMK25: Addressed in my latest commit.

src/player/player.controller.ts
CapoMK25: Addressed in my latest commit.

hoan301298:
> @hoan301298 or @Iir0R you can check and approve once more before I hit merge on this, thanks!

I checked your changes, looks good to me

PR #778, opened 2026-02-12
777 Bug chat messages fail validation when sent to the server

CapoMK25:
### Brief description

A new bug issue emerged from the UI team where the chat messages failed the validation when being sent to the server. 

Turns out the the `ValidationPipe` was colliding with the internal data injection on the server side. While the Unity client was sending simple strings, the server-side logic was injecting the `sender_id` and `clan_id` as MongoDB `ObjectId` instances before the validation phase. Since the `CreateChatMessageDto` was strictly looking for strings via `@IsMongoId()`, the validator threw an `unknownValue` error upon encountering the MongoDB objects.

### Change list

- Resolved DTO Type Collision: Updated `CreateMessageDto` to use the `@Allow()` decorator for` sender_id` and `clan_id`. This allows the validator to accept both hex strings from the client and `ObjectId `instances injected by the server's internal "decorators"/"guards".

- WebSocket Stability: Fixed the `clanMessage` event failure that was preventing players from sending messages in the game client.

- Validation Logic Verification: Added a temporary unit test to simulate the server's internal state (validating DTOs with real `ObjectId` instances) to ensure nothing bad occurs when IDs are pre-populated. This test passed with flying colors but has been excluded and removed from this PR. 

## Technical Notes

The root cause was identified to be a mismatch between the transport layer (stringified JSON) and the application layer (hydrated MongoDB objects). By loosening the DTO constraints to allow both types regarding the ID format, we maintain security while allowing the NestJS lifecycle to complete its natural flow.


hoan301298 (changes requested):
I don’t think we should use @Allow() for these ID fields. It effectively bypasses validation and allows any value to pass, even though these fields are supposed to contain valid Mongo IDs. This removes an important safety check and makes the server more vulnerable to invalid data. Instead, we should use a validator that explicitly accepts valid Mongo ID strings (and ObjectId instances if needed), so we preserve strict validation while still supporting our server-side injection flow.

I suggest to create a new validator and implement it into these ID fields that suppose to be a valid ObjectId or Mongo ID string:
```typescript
import { registerDecorator, ValidationOptions } from 'class-validator';
import { ObjectId } from 'mongodb';

export function IsMongoIdOrObjectId(options?: ValidationOptions) {
  return function (object: any, propertyName: string) {
    registerDecorator({
      name: 'isMongoIdOrObjectId',
      target: object.constructor,
      propertyName,
      options,
      validator: {
        validate(value: unknown) {
          if (value instanceof ObjectId) return true;

          return (
            typeof value === 'string' &&
            ObjectId.isValid(value)
          );
        },
      },
    });
  };
}
```

Then implementation:
```typescript
@IsNotEmpty()
@IsMongoIdOrObjectId()
sender_id: string | ObjectId;

@ValidateIf(o => o.type === ChatType.CLAN)
@IsMongoIdOrObjectId()
clan_id?: string | ObjectId;
```

Let's try to test with this solution.

CapoMK25:
I forgot to comment earlier. The solution has now been implemented correctly and tested + verified on unit tests and the classic Postman API tests. 


Postman_PATCH

hoan301298 (commented):
I think your solution is good in logic from now and we might need to test this solution via `WebSocket` due to the issue caused from it. Like we had the issue on PR - 755. Tests by requesting API on Postman seem to be not sufficient in this case.

@BillTheBeast Hi, if you can guide us how to test with `WebSocket`, that would be awesome and we can move faster for bugs in chat/WebSocket. Maybe using Eclipse Mosquitto?

**UPDATED**

I already fixed the problem and knew how to test with WebSocket and all the tests passed via WebSocket on Postman. This PR is now good to merge into dev.

hoan301298 (approved):
The problem was from the last PR that related to chat module when adding transaction support. 

The error is from `validate()` function, this function validate only the class and not plan object. Check the example below:

```typescript
// This is plain object
const chatMessage: CreateChatMessageDto = {
  ...
}
```

```typescript
// This is class for dto
const chatMessage = new CreateChatMessageDto ({
  ...
})
```

BillTheBeast:
> @BillTheBeast Hi, if you can guide us how to test with `WebSocket`, that would be awesome and we can move faster for bugs in chat/WebSocket. Maybe using Eclipse Mosquitto?

I don't really have a good suggestion for how to test the code other than determining the format you want the json object to be extracted in and then testing if the data in that object is correct. At least to my knowledge, Jeri was doing all of his testing with Postman and while I've mostly just tried sending data to the endpoint from the game client itself.

This is the basic format of turning the json text into something usable in C#: (Using section from the customCharacter POST endpoint as an example)
JObject result = JObject.Parse(request.downloadHandler.text); (request.downloadHandler.text is the json from server)
ServerCharacter serverCharacter = result["data"]["CustomCharacter"].ToObject<ServerCharacter>();

hoan301298:
Hi @BillTheBeast 

I have solved this problem and already tested the solution via Postman. Now everything should work as expected.

Today we will finish this issue and the 2nd message was sent back. I have troubleshooted it and tested it. Two of them should be merged into dev branch today and closed.

Thank you very much.

PR #780, opened 2026-02-17
Implemented tested endpoints

### LINKED ISSUE #759: Feature request: Update the battle data handling and endpoints.

Endpoints and the background validation should be overhauled to make sure that the battle data is properly handled  by the server.

Firstly: There should be an endpoint where the client can inform the server that a battle is starting and to register it with the server.

POST: /gamedata/battle/start (Can also use the preexisting "/gamedata/battle/" endpoint you would rather use the "type" field to split the register and result)

**This should take the following information.**
Gametype: Enum_string (matchmaking, casual, custom)
Team1: [player_id, player_id] (one id is needed the other can be empty.)
Team2: [player_id, player_id] (one id is needed the other can be empty.)
MatchID: string_id (optional if not given then server generates one.)

Return value should include the MatchID, should the registering be successful.

Secondly: When the match finishes each player sends the result to the server.

POST: /gamedata/battle/result (Can also use the preexisting "/gamedata/battle/" endpoint you would rather use the "type" field to split the register and result)

**This should take the following information.**
MatchID: string_id (should match the registered id)
Duration: string or float in seconds.
Result: int (1 if team 1 won, 2 if team 2 won)
Senders ID can be determined by the AccessToken sent with the message.

Return value should include the token for raid.

By registering the data at battle start it is easier to validate that a battle actually happened, of what type it is and who should be sending the result for that battle in a minute or two.

Then when the players send the results to server these can be checked to see if they match and if someone doesn't send one we can assume that something went wrong somewhere but we can still use the data from the other players to distribute the potential rewards and log the battle results.
And if the results don't match we can potentially check which result is the odd man out and disregard it.

If match is registered but no results are received (30 mins timeout) then the match is closed and discarded.
When all player have sent their results the match the match is closed and the results are handled.
If at least two players send their results a 2 min timeout is started after which the match is closed and the results are handled.
If the results are vastly different (all player send different information) we can assume that something went wrong and we can discard the results.

If the player is being registered for another match while they haven't sent the results for the previous one we can assume that they aren't sending any for the previous one.

## Any additional info?

The timeouts and validation don't need to fully be there for the first iteration as the data sent would be from internal testing anyway and thus not 100% mission critical yet.

CapoMK25:
### Brief description

This PR implements a robust state machine for handling battle results in the `BattleService.` It introduces logic to handle "Odd Man Out" scenarios where players report conflicting results (e.g., Team 1 claims victory while Team 2 also claims victory).

The system now transitions from `OPEN` to `PROCESSING` when results conflict, triggering a 2-minute "Final Call" timer. After the timeout, the service resolves the conflict via a majority vote (defaulting to Team 1 in ties) and distributes Raid Tokens. This ensures the integrity of the game economy and prevents "silent agreements" or stuck matches.

### Change list

- BattleService: Implemented `handleBattleResult` with conflict detection and `resolveConflict` with a majority-vote algorithm.

State Machine: Integrated `BattleStatus.PROCESSING` to represent matches awaiting a timeout resolution.

Timer Logic: Added a 2-minute `setTimeout` mechanism to finalize matches that don't reach a natural consensus.

Raid Token Distribution: Refactored `generateRaidTokens` into a dedicated private method to ensure rewards are only issued upon final verification.

Data & Documentation

Schema Update: Updated `BattleSchema` with pre-validation middleware to prevent players from being assigned to multiple teams.

JSDocs: Added comprehensive documentation for the `BattleService`, `Battle Schema`, `BattleStatus` Enum, and associated DTOs.

DTO Alignment: Fixed a property mismatch where result was used in the DTO but expected as `winnerTeam` in the service.

Testing & Quality

Unit Tests: Created `battle.service.test.ts` using Jest fake timers to verify the full lifecycle of a conflict resolution.

Linting: Resolved ESLint errors regarding empty blocks and unused variables to pass Husky pre-commit hooks.


hoan301298 (changes requested):
I think you were quite wrong to create a new module `battle` since our `gameData` module was definitely used for this issue.

To-do steps:
- Remove `battle` module
- Integrate the data in the issue description from Niko to modify `game.schema.ts`.
- Keep `gameData.controller` with `@POST('battle/start')` to register or start the battle (auto-created `matchId` as `_id` if not existed, gameType, ...)
- Create new API endpoint with `@PUT('battle/result')` to let client sends result of the battle and update the existed battle data.

CapoMK25:
Swagger to be updated with the new endpoint data later on.

CapoMK25:
Looks like I messed up the Git branches on this issue. I'm looking into this tonight.

CapoMK25:
Jenkins is blocked completely and pending on access to VPS from my side. I've now resolved this PR's changes though.

CapoMK25:
@hoan301298 I think you can review this when you have the time again. Looks like the codeowner change only applied to new PRs and not this one.

CapoMK25:
@Iir0R You also have admin rights to walk through this PR now that Hoan is more absent.

hoan301298 (approved):
Looks good! Ready to merge @CapoMK25

Issue #741, opened 2025-12-15
Custom character stat storing


PlayJeri:
#### Problem
Currently custom character stats are stored as absolute values instead of deltas from base stats.

#### Current behavior
- Each character document contains stats like hp, attack, shield, speed.
- These stats are stored as numeric levels
- Every character has predefined base stat levels. 
- When a character is created, the base stat levels are saved directly in DB.
- When player levels up a character, the updated stat levels overwrite these values in the DB.

#### Issue
Because the database stores final stat levels, we lost the distinction between:
- base stats defined by game balance
- stats gained through player progression

If we later need to balance a character base stats:
- We don't know which portion of the stored values comes from the base stats vs. player level-ups.
- Updating base stats would require complex migrations or would risk breaking player progression.

#### Solution
Store stat level-ups as deltas from base stats.
Instead of storing absolute stat levels in the database, store only the level-up amounts (deltas) relative to the character's base stats. Then end points that return the custom character stats / levels you return the saved level-ups from db + the character base stats.

### Tasks

- [ ] Check if the schemas / DTOs related to character stats need to be updated (documentation or code itself)
- [ ] Update service methods that return character stats to return base stats + stored level ups.

CapoMK25:
Closing this as the PR has been merged into dev.

PR #781, opened 2026-02-21
741 custom character stat storing

Iir0R:
### Brief description
Updated customCharacter.service to meet required changes in #741 Custom character stat storing. Also changes to tests.

### Change list

- New function: addValues, gets document as POJO, adds matching base stats, returns result.
- New function: getFieldsToUpdate, adds key value pairs to correct key, $inc or $set.
- createOne creates a new character with stat deltas starting at 0, returns base + deltas.
- readOneById, readOne, readMany and readPlayerBattleCharacters use addValues, return base + deltas.
- updateOneById and updateOneByCondition use getFieldsToUpdate to increment deltas and set others.
- Changes to __tests__ custom character files to match new results.

### DIFF
--- a/src/__tests__/player/customCharacter/CustomCharacterService/readMany.test.ts
+++ b/src/__tests__/player/customCharacter/CustomCharacterService/readMany.test.ts
@@ -57,12 +57,22 @@ describe('CustomCharacterService.readMany() test suite', () => {
 
     const clearedCharacters = clearDBRespDefaultFields(characters);
 
+    const expected1 = clearDBRespDefaultFields(
+      characterService.addValues(character1),
+    );
+    const expected2 = clearDBRespDefaultFields(
+      characterService.addValues(character2),
+    );
+    const expected3 = clearDBRespDefaultFields(
+      characterService.addValues(character3),
+    );
+
     expect(errors).toBeNull();
     expect(clearedCharacters).toEqual(
       expect.arrayContaining([
-        expect.objectContaining({ ...character1 }),
-        expect.objectContaining({ ...character2 }),
-        expect.objectContaining({ ...character3 }),
+        expect.objectContaining(expected1),
+        expect.objectContaining(expected2),
+        expect.objectContaining(expected3),
       ]),
     );
   });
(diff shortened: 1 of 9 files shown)


hoan301298 (commented):
Sorry for this inconvenience, but this PR has many changes and need to be tested later on. I am still reviewing it. Thanks

src/__tests__/player/customCharacter/CustomCharacterService/readMany.test.ts
```diff
@@ -72,12 +82,22 @@ describe('CustomCharacterService.readMany() test suite', () => {
       filter: undefined,
     });
 
+    const expected1 = clearDBRespDefaultFields(
```
CapoMK25, line 85: This is actually ok here, but in the future if possible, you could try to integrate into a solution where there's less repetition if that's possible. But overall I'm approving this regardless as I was guilty of the same stuff before AI in 2021.

CapoMK25 (approved):
Looks good to me, and since this works in unit testing as well as on Postman, this should be good to go from my eyes.

hoan301298 (changes requested):
These results I've gotten below by calling APIs in Postman respectively with:

- `POST` - create a new character -> 201 status_code (success)

```typescript
"CustomCharacter": {
  "_id": "699c17d246f1ba5be0c956a0",
  "characterId": "201",
  "defence": 5,
  "hp": 5,
  "size": 4,
  "speed": 8,
  "attack": 8,
  "level": 1,
  "player_id": "6990a77342de2a35eedcdfc6"
}
```

- `PUT`  - update character stats  -> 204 status_code (success)

```typescript
{
  "_id": "699c17d246f1ba5be0c956a0",
  "characterId": "201",
  "defence": 30,
  "hp": 120,
  "size": 3,
  "attack": 50,
  "speed": 6,
  "level": 5
}
```

- `GET` - get character stats after update -> 200 status_code (success)

```typescript
{
  "_id": "699c17d246f1ba5be0c956a0",
  "characterId": "201",
  "defence": 35,
  "hp": 125,
  "size": 7,
  "speed": 14,
  "attack": 58,
  "level": 5,
  "player_id": "6990a77342de2a35eedcdfc6"
}
```

If you can see the results above. So technically the GET result was wrong by logic. By following in description of issue #741:

-  With the UPDATE data, we actually got the updated stats so the deltas should be store as: `deltas` = `updatedStats` - `baseStats`. Therefore the function `getFieldsToUpdate` should be fixed by logic.

--> The correct version: `GET` stats equals to `UPDATED` stats

hoan301298:
Anyway, you did very good in this issue. Definitely this issue is not easy to start with, but I can see here that you deeply understand the issue, where to implement. Just small logic needs to be fixed.

Well done!

hoan301298 (approved):
The solution already tested on Postman!

Well done!

Issue #782, opened 2026-02-23
Bug: Unable to update playerData into the server.


BillTheBeast:
## What doesn't work?

When trying to send a updated playerdata to the server, it returns 400 Bad Request error.

## What is the endpoint and method? (if applicable)

Endpoint:  player/

Method: PUT

## How can it be reproduced?

1. Send a message to the https://devapi.altzone.fi/player/ 
i.e.
{
  "_id": "69066ef167be4c41fdcde73a",
  "name": "BTB3",
  "clan_Id": "6740af56d977418ddbe08e29",
  "avatar": {
    "head": 0,
    "hair": 1030103,
    "eyes": 2140102,
    "nose": 2240101,
    "mouth": 2370101,
    "eyebrows": 0,
    "clothes": 3140101,
    "feet": 3310101,
    "hands": 3260101,
    "skinColor": "#FFFFFFFF",
    "hairColor": "#FFFFFFFF",
    "eyesColor": "#FFFFFFFF",
    "noseColor": "#FFFFFFFF",
    "mouthColor": "#FFFFFFFF",
    "clothesColor": "#FFFFFFFF",
    "feetColor": "#FFFFFFFF",
    "handsColor": "#FFFFFFFF"
  },
  "currentAvatarId": 701,
  "battleCharacter_ids": [
    "6930cc09437dd6daf71635ef",
    "6930cc09437dd6daf71635f7",
    "691462ebd014d2f595a258ef"
  ]
}

## What should it do instead?

As the structure of the json hasn't been changed in clientside and no changes to the API have been reported, there shouldn't be issues accepting it.

## Anything else you want to add?

If something has been changed in the api itself, it should be updated to the swagger.

CapoMK25:
This one is very directly related to #785 so I'm waiting on that to be merged before addressing this one in order to avoid conflicts.

BillTheBeast:
Yeah, I read that one. If the format that the data needs to be given has changed then I just need to know what the new format is. If what I wrote above should work then I'll leave you to see what the problem is.

CapoMK25:
With the PlayerAvatar PR change, it was changed to be an object from a flat number through a schema change. The old way was basically this: 

```
"head": 0,
"hair": 1030103,
"hairColor": "#FFFFFFFF"
```

And the new way is basically like this: 

```
"head": { "id": 0, "color": "#FFFFFFFF" },
"hair": { "id": 1030103, "color": "#FFFFFFFF" }
```

[...]

PR #785, opened 2026-02-23
783 Error refactored

CapoMK25:
### Brief description

Looks like when sending the older JSON input to Postman on the PUT player endpoint it ran into this issue: 

`Error putting data to https://devapi.altzone.fi/player/ - HTTP/1.1 400 Bad Request: {"statusCode":400,"error":"Bad Request","errors":[]}`

The error message setup had to be refactored to counter this for good, since in its current form it just didn't accept the old format before the change when the playerAvatar was updated to an object, but instead of giving a proper error message explaining why, it just went quiet. This has now been addressed with a proper error message explanation. Here's the message:

postman_output_JSON

### Change list

- `main.ts`: Deleted the currently unused imports. 
- Refactored the `const` variable with a helper function and a lot more stuff to give the user a better output when the wrong `JSON` is the input. 
- Added a general message "`Validation failed`" in the bottom of the function. 

Tested on Postman too and the output can be seen on the image.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -2,8 +2,6 @@ import { NestFactory } from '@nestjs/core';
 import { AppModule } from './app.module';
 import { BadRequestException, ValidationPipe } from '@nestjs/common';
 import { useContainer, ValidationError } from 'class-validator';
-import { APIError } from './common/controller/APIError';
-import { validationToAPIErrors } from './common/exceptionFilter/ValidationExceptionFilter';
 import EnvHandler from './common/service/envHandler/envHandler';
 import cookieParser from 'cookie-parser';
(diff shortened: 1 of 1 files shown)


hoan301298 (changes requested):
Since Niko, UI Team already knew the bug was the mismatch of their requests, I believe this PR and the PR for issue #782 should be closed. Maybe we can do another PR for clearing info of AvatarPlayer of PUT method

CapoMK25:
Closing per Hoan's request. The UI mismatch was resolved on the client side by the UI Team. Moving the remaining avatar clearing' logic to a fresh PR to keep things focused.

PR #786, opened 2026-02-24
Configured migrate-mongo as the DB migrations tool

### LINKED ISSUE #451: Configure DB migrations tool
### General description

There should be created a setup for mongo migrations.

### Things to notice

The tool for migrations should be easy to use in a team.

The mongoose automatic migrations must be disabled.

### Tasks

- [ ] Add mongo migration tool

CapoMK25:
### Brief description

Mikhail as the previous Server Team Lead opened an issue exactly one year ago: #451

I recognized when working on the server that this was the first nice-to-have change and implementation that was actually needed to be done, since the DB was always out of whack with any larger Schema changes that have become increasingly common with recent issues and PRs on the backend. 

Setting the indexing off with the `autoIndex` also attempts to tackle this, as this was mentioned on the issue too.
 
After looking online, I stumbled upon `migrate-mongo` as the one and only MVP solution to tackle this problem.

Note: I included the `database/migrations` directory with a `.gitkeep` file. This ensures the folder structure exists in the repo so that `migrate-mongo` commands work immediately upon pulling, while keeping the directory clean of any test data.

Here's a rough draft of how this should work from here on out whenever you're working with this tool:
Use these four commands:

Create a new change: `npm run migrate:create <name-of-change>`
This creates a timestamped file in `database/migrations` based on the template.

Apply pending changes: `npm run migrate:up`
Run this after pulling new code to sync your local DB.

Check migration status: `npm run migrate:status`
See what has been applied and what is pending.

Undo the last change: `npm run migrate:down`
Use this if you made a mistake or need/want to revert a schema change.

### Config

`migrate-mongo-config.js` is at the project root and uses an `.env` you have locally

Run all commands from the project root (where `package.json` lives), not from the `database/ subdirectory`.

### Change list

-`migrate-mongo-config.js` — baseline config at project root, loads `.env` via `require('dotenv').config()`
-`package.json` — includes the four `migrate:* npm` scripts
-`app.module.ts` — automatic index creation disabled (`autoIndex: false`)
-`database/migrations/` — directory with `.gitkeep` for version-controlled folder structure
-`migration-template.js` — CommonJS template for new migrations (`module.exports.up/down`)


src/app.module.ts
```diff
@@ -70,6 +70,7 @@ const authGuardClassToUse = isTestingSession() ? BoxAuthGuard : AuthGuard;
         return {
           uri: mongoString,
           dbName,
+          autoIndex: false,
```
hoan301298: why is this false? `autoIndex` is true by default and helps us not to create a new index manually.

migration-template.js
hoan301298, line 1: Those arrow functions were not accomplished?

hoan301298 (commented):
I see a few issues in approving this PR. Database migration is extremely complex.

I think we need consensus and especially, appropriate documentation to finalize this.

migration-template.js
CapoMK25: The arrow functions here were meant to be just initial placeholders. I wrote/generated custom template tests on these previously when testing out the migration in practice, but decided to leave the template file as-is for this actual PR since the migrations would very rarely be applicable to a "one-size fits all" approach, and would need to be adjusted during each migration here on this template. But like I said, this is just a placeholder template for now and the final approach is still up for debate of course.

src/app.module.ts
CapoMK25: Agreed, false could've worked for a bigger game backend to avoid any race conditions, but for the scope of AltZone, I'm reverting this back to true for now.

CapoMK25:
> I see a few issues in approving this PR. Database migration is extremely complex.
> 
> I think we need consensus and especially, appropriate documentation to finalize this.


I agree 100%, but I'm still unsure of where the best place for this would be. I think the internal wiki and the README could work? My apologies, I just wanted for everyone else to see my approach and take on the migrations so I pushed this PR out the door first before taking documentation truly into account for the whole team.

CapoMK25:
This is not a priority at all.

tickBit:
There seems to be a small conflict regarding lodash module (current dev branch has later version). I looked locally the matter, and apparently I have different npm version (mine is 11.12.1), because in order to avoid garbage in the package-lock.json, the project should be re-build. In my local experiment the package-lock.json file was a bit more changed, than expected. So to avoid more merge conflicts, I leave this matter for now.

CapoMK25:
@tickBit I'll let you have the final say on this whenever this becomes relevant again in the future. Currently this is only a nice-to-have feature and doesn't impact a whole lot, would only maybe make the team's lives a little simpler and easier if implemented well. My original point still stands, I'd like to keep this PR in place for as long as possible as a placeholder since the boilerplate setup has already been implemented, but this should not be a priority at all right now, or even in the future.

hoan301298:
Hey, sorry for the confusion on my end, a lot has changed since this PR was raised, with many new features coming in and the DB schema evolving quite a bit.

Regarding your instructions for the migration tool: could you move all the migration files into a `database/` directory from the project root instead of `src/`? It would also be great to have a script file to wire up those `npm` scripts.

Correct me if I'm wrong, but the workflow would be: make the logical changes to the edited model (in the template) → run `migrate:up` to apply them → verify the data looks correct after applying.

Really good idea overall. To make it clearer for the team, it would help to also add a `markdown` file in that same `database/` folder covering the workflow and common use cases.

@CapoMK25 @tickBit  any thoughts?

migrate-mongo-config.js
```diff
@@ -0,0 +1,27 @@
...
+const url = `mongodb://${MONGO_USERNAME}:${MONGO_PASSWORD}@${MONGO_HOST}:${MONGO_PORT}/?replicaSet=rs0`;
+
+const config = {
+  mongodb: {
+    url: url,
+    databaseName: MONGO_DB_NAME,
+    options: {
+    }
+  },
+  migrationsDir: "src/database/migrations",
```
tickBit: The migration files could be in `/database` instead of `src/` as @hoan301298 pointed out.

tickBit (changes requested):
Please put migration files in `database` instead of `src` and add also markdown documentation how to use the database migration into `database` folder.

Really useful utility particularly now, when there will be database changes in quite near future in more than 1 feature requests.

CapoMK25:
@tickBit I saw the need for this on Discord. I'm continuing with this fully by tomorrow by the latest since the original implementation was 6 months ago when this PR was opened, and I honestly don't remember even half of the original idea anymore.

CapoMK25:
Readme has been added as the baseline placeholder documentation.

CapoMK25:
To be tested out before merging by me on the audio side specifically if possible.

tickBit (approved):
Lots of thought has been given to this. Good work! Approved.

tickBit:
I would just like to double check, that is in `app.module.ts` `autoIndex: true` intentional? If it is, this is ready to be merged.

CapoMK25:
autoIndex should be on true as it is right now, I flipped it to false on an false account months ago. It was originally on true, and doesn't affect the implementation.

PR #787, opened 2026-02-24
784 Changed message limit from 64 to 256

### LINKED ISSUE #784: Feature request: Increase the chat message character limit.
## What is the problem?

Some of the latest changes added a 64-character limit to the messages that can be sent through chat. This is extremely short and is causing problems currently and even if we would change the structure a little, it will become a problem if we decide to allow free communication over the chat in the future. 

Increasing the limit to at least 200 (or 256 if the power-of-2 should be maintained.) should be sufficient for most messages. (128 is likely still a little too low)

Iir0R:
### Brief description
Changed message character limit from 64 to 256

### Change list

- Changed createMessage.dto.ts content MaxLength from 64 to 256
- Updated comment

### DIFF
--- a/src/chat/dto/createMessage.dto.ts
+++ b/src/chat/dto/createMessage.dto.ts
@@ -34,13 +34,13 @@ export class CreateChatMessageDto {
 
   /**
    * Text content of the message.
-   * Content must not exceed 64 characters
+   * Content must not exceed 256 characters
    *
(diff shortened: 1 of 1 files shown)


hoan301298 (approved):
Good job!

CapoMK25 (approved):
Looks very good to me.

PR #801, opened 2026-04-03
Codeowners changed

CapoMK25:
### Brief description

As I'll have to step back from the server starting next week and reduce my hours to only some evenings and weekends, this code change is necessary going forward with the server. This has been discussed with psykkis as well and this change grants codeowner privileges to two crucial members.

### Change list

- Added @Iir0R and @psykkis as codeowners
- Removed @jkrizsan from codeowners
- @PlayJeri can and should stay for now just in case


CapoMK25:
@hoan301298 No rush but this is a clean PR to approve and merge from here when you can, thanks!

hoan301298 (approved):

PR #805, opened 2026-04-12
updated README to address env and fixed the Number mismatch 

### LINKED ISSUE #804: Update the README with info about the .env files for both Windows and Linux setups
### Links

[Swagger]()

[Flow chart]()

### General description

Currently the `README.md` is missing the `.env` completely from everywhere. This is more than required when setting up the environment for the first time, `npm run start:dev` literally won't work without it at all.

### Things to notice

I'm not sure whether or not the `README.md` should include the .`env` file itself or not, but at least the `README.md` should be updated with information regarding how to set it up for the first time locally at the root of the project.
 

### Tasks

- [Update `README.MD` with info on how to set up the `.env ` for a new user.]

CapoMK25:
### Brief description

The `README.md` documentation never mentioned the need for an `.env` file, causing problems for new starters on this repo due to it missing in their configuration. I decided to add that into the documentation briefly. The `.env` pasted here is taken from my own Linux workstation and should be modified with the secrets when taken into use on Linux. Windows has its own `.env` file, but I felt like pasting that in here as well would've made this whole thing 10x longer than it can realistically aim to be. 

I also ran into husky failing with 2 errors regarding the Number lowercase vs uppercase config on 2 separate files. Even though it's unrelated, this popped up suddenly on the dev branch and was reported by the new lead Markus as well. So I figured I'd combine it here.

### Change list

- Changed `README.md`
- Changed the tokenversion from uppercase to lowercase on both `src/profile/profile.schema.ts` and `src/profile/profile.service.ts`


README.md
hoan301298, line 1: We should never mention the `.env` file here and this project is public, not private

hoan301298 (changes requested):

README.md
CapoMK25: My bad! I'll modify the README with another commit to take that out but I'll mention that any new starter will need to create an .env file themselves.

CapoMK25:
@Iir0R When you have the time, you can also walk through this PR by the looks of it, thanks!

init-db-on-windows.ps1
```diff
@@ -10,7 +10,7 @@ $USERNAME = "rootUser"
 $PASSWORD = "superSecretPassword"
 $DATABASE = "altzone_dev"
 $REPLICA_NAME = "rs0"
-$REPLICA_HOST = "localhost"
+$REPLICA_HOST = "altzone_db_dev"
```
tickBit: I recently set up again my backend system and with `$REPLICA_HOST = "localhost"` it worked.

tickBit (changes requested):
I think, it's safest to use `$REPLICA_HOST = "localhost"` because it worked at least on my computer, when I recently set up the DB again.

init-db-on-windows.ps1
CapoMK25: Done on my latest commit.

tickBit (approved):
Looks ok. Let's see, if my approval frees the merging button now.

CapoMK25:
@hoan301298 I've addressed the feedback but looks like this still requires your action. Could you please re-review when you have a moment? Thanks!

PR #810, opened 2026-04-14
Fix issue 808 mongo db init script and linting problems

tickBit:
### Brief description

Fixed PowerShell script to initialize MongoDB on Windows.
Fixed linting problems: Number vs. number changed to number.

### Change list

- `init-db-on-windows.ps1`
- `src/profile/profile.schema.ts` and `src/profile/profile.service.ts`
- some auto corrections, I think: I touched on those 3 files only


init-db-on-windows.ps1
```diff
@@ -10,7 +10,7 @@ $USERNAME = "rootUser"
 $PASSWORD = "superSecretPassword"
 $DATABASE = "altzone_dev"
 $REPLICA_NAME = "rs0"
-$REPLICA_HOST = "altzone_db_dev"
+$REPLICA_HOST = "localhost"
```
CapoMK25, line 13: Check this. On this commit you're changing the `$REPLICA_HOST` to `localhost` but on the other commit it's doing the opposite and changing it to `altzone_db_dev`. Whichever one works should be the one to stick with going forward.

CapoMK25 (changes requested):
All good except for the localhost vs altzone_db_dev mismatch. In terms of the ESLint changes, they're always a headache and Hoan hates seeing them scattered around PRs for understandable reasons of cluttering the commits and the code. Thus I recommend in the future to come up with a more comprehensive strategy on what to do with all the forced linter changes on every commit since they're baked in by default to every single commit you do.

init-db-on-windows.ps1
tickBit: At the moment, I have `$REPLICA_HOST = "localhost"` and the server works and MongoDB gets set up.

I had earlier problems with setting up MongoDB, I asked ChatGPT and this was the answer :) .

CapoMK25:
This should be merged into dev ahead of #805 to avoid any merge conflicts. I'll resolve that PR once this is merged into dev. I tried cherry picking the best parts from this PR to get past husky on that PR but it looks like it brought the script changes with it from the same commit, naturally.

init-db-on-windows.ps1
CapoMK25: So it should be `localhost` instead of `altzone_db_dev` in order for the script to work 90% of the time from here on out for other possible starters in the future as well? If the mostly working one is `localhost` I naturally would require to keep this commit, but get rid of the earlier `altzone_db_dev` commit/code snippet since the host has to be either this or that.

tickBit: As far as I can see, `localhost` is the way to go.

CapoMK25 (approved):
Approved, only the linter changes should again be figured out later. Good job.

PR #820, opened 2026-04-29
Hotfix regarding issue 816

tickBit:
### Brief description

For completeness and to full fill client side's requirements also furniture IDs are now returned. They tell indirectly to which stall the furniture belongs to (GET /stall).

Example object:
{
  "adPoster": { "border": "solid", "colour": "blue" },
  "maxSlots": 7,
  "furnitureItemIds": ["507f1f77bcf86cd799440011", "507f1f77bcf86cd799440022"],
  "furnitureItems": ["Chair_Rakkaus", "Table_Lamppu"]
}

I forgot in the previous PR to uncomment one line related to testing. Now it's fixed.

For some reason from file `clanShop.service.ts` line `return endTransaction(session, true);` was missing, even though I hadn't touched that file. I committed the missing line, too.

GET /stall/ownClan returns furniture items from basicService.

### Change list

- among other files also `src\fleaMarket\stall\dto\stallResponse.dto` was changed so, that furniture IDs are returned, they indirectly refer to the stall the furniture belong to

### DIFF
--- a/src/fleaMarket/stall/dto/stallResponse.dto.ts
+++ b/src/fleaMarket/stall/dto/stallResponse.dto.ts
@@ -48,13 +48,14 @@ export class StallResponse {
   maxSlots?: number;
 
   /**
-   * ID of the stall's furniture items
-   * The ID is indirectly related to the stall the furniture item is in
-   * @example "507f1f77bcf86cd799440011"
+   * IDs of the stall's furniture items
+   * These IDs are related to the furniture items in the stall and can
+   * be used to fetch more detailed information about the items if needed.
+   * @example ["507f1f77bcf86cd799440011", "507f1f77bcf86cd799440012"]
    */
   @Expose()
-  _id?: string;
-  
+  furnitureItemIds?: string[];
+
   /**
    * Furniture items that are currently in the stall
    * @example ["chair", "lamp"]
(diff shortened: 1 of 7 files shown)


tickBit:
All green now. I re-ran the tests. In the first run there was that usual duplicate DB key problem.

tickBit:
I added one more condition, which is what I think, the client side has meant in fetching the furniture from /stall endpoint.

All the test are green. Now tests were ran only once.

src/fleaMarket/stall/dto/stallResponse.dto.ts
```diff
@@ -47,6 +47,14 @@ export class StallResponse {
   @Expose()
   maxSlots?: number;
 
+  /**
+   * ID of the stall's furniture items
+   * The ID is indirectly related to the stall the furniture item is in
+   * @example "507f1f77bcf86cd799440011"
+   */
+  @Expose()
+  _id?: string;
```
CapoMK25, line 56: How come the `id` is partly optional? If every ID should usually be unique and unchanged (at least in SQL  and IT in general)? Nitpicking with this comment but I'm genuinely curious, especially since TS can be a PITA when it comes to not making, or making, these IDs and other variables optional.

CapoMK25 (approved):
Good job again! Approved, I left my only comment as an observation before noticing that the ? was used even earlier in the code to begin with, thus I won't comment on it further.

Issue #821, opened 2026-05-05
Bug: Getting player data with DailyTask throws 403


tickBit:
## What doesn't work?

When player has signed in, has token, GET /player/{_id}?with=DailyTask throws ` 403 Forbidden: {"message":"The logged-in user has no permission to execute read_request action","errors":[{"response":"NOT_AUTHORIZED","status":403,"message":"The logged-in user has no permission to execute read_request action","name":"APIError","reason":"NOT_AUTHORIZED","field":null,"value":null,"additional":null,"statusCode":403}],"statusCode":403,"error":"Forbidden"}`

## What is the endpoint and method? (if applicable)

Endpoint: 

`/player/{_id}?with=DailyTask`

Method: 

GET

## How can it be reproduced?

1. By calling GET `/player/{_id}?with=DailyTask`

## What should it do instead?

The call should return a `Player` object. Example:

```
{
    "data": {
        "Player": {
            "_id": "69f98622e122b244d6f0d49c",
            "name": "Guest-2e5fe052",
            "points": 0,
            "battlePoints": 0,
            "backpackCapacity": 0,
            "uniqueIdentifier": "Guest-2e5fe052",
            "above13": null,
            "parentalAuth": null,
            "battleCharacter_ids": [],
            "currentAvatarId": null,
            "profile_id": "69f98621e122b244d6f0d49a",
            "clan_id": "69ef8b7332f46bd743f4c558",
            "DailyTask": null,
            "avatar": null,
            "clanRole_id": "69ef8b7332f46bd743f4c55a",
            "createdAt": "2026-05-05T05:54:42.000Z",
            "playstyle": "Balanced",
            "carbonFootprint": 0,
            "clanCoinsAccumulated": 0,
            "emotions": []
        }
    },
    "metaData": {
        "dataKey": "Player",
        "modelName": "Player",
        "dataType": "Object",
        "dataCount": 1
    }
}
```

tickBit:
Fixed.

PR #822, opened 2026-05-05
Switch @AddType('PlayerDto') to class PlayerDto

tickBit:
### Brief description

Small bug fix for #821 . One decorator switched to correct place.

For future reference: If `export class StatDetailDto` ever needs a decorator, it should be `@AddType('StatDetailDto')`.

### Change list

- `player.dto.ts`


Iir0R (approved):
Looks good to me, approving.

PR #797, opened 2026-03-18
Implemented clan changes happening through a vote

CapoMK25:
### Brief description

The UI Team asked to heavily prioritize issue #651. Here's the very heavy refactor/implementation end product for that.

This has been tested as working on Postman and on unit tests as well and implemented to follow best practices (except the `as any `in tests). I attempted to follow the existing functionality and style religiously, even though the refactor/implementation was huge, error-prone and took an insane amount of time to get finished (unit tests passing, Postman and MongoSH working as intended). 

Note: I'm fully aware `as any` is used in the tests here. This is just so that I could get the tests to pass locally, since `as any` is used in unit tests in other files as well, despite it being bad practice in the actual operational code side of things. 

### Change list

- Added new file `src/clan/interface/clanGovernanceUpdate.interface.ts` to house the interface instead of it being on the DTO file previously
- Added a new file `src/gameEventsEmitter/payloads/UpdatedClan.ts` to follow the previous logic on `event.types.ts`
- Added a new file `src/voting/type/governancePayload.ts` to be the main functionality of this implementation
- Refactored the tests to address this massive refactored implementation following an MVP style approach
- Refactored the `src/clan/clan.controller.ts` file briefly to address the changed endpoint
- Refactored the `src/clan/clan.service.ts` file MASSIVELY to support the mandated new voting functionality in tandem with voting itself
- Imported DTOs and the GovernancePayload + changed existing functionality to meet the issue requirements on `src/clan/dto/updateClan.dto.ts`
- Readded JSDocs back to `src/clan/clan.service.ts` with improved ones layered in the middle
- Followed the existing setup on `src/gameEventsEmitter/event.types.ts` (mentioned here again)
- Imported and added the voting-related GovernancePayload on `src/voting/dto/createVoting.dto.ts`
- Same thing on `src/voting/dto/voting.dto.ts`
- Added the required new enum field to `src/voting/enum/VotingType.enum.ts`
- Modified the schema at `src/voting/schemas/voting.schema.ts` to add the exact same new GovernancePayloadVoting there as well
- Added the GovernancePayload to `src/voting/type/startItemVoting.type.ts` as well (as it was required)
- Added the same "ClanGovernance" functionality to `src/voting/voting.module.ts` as well

issue631

PR #830, opened 2026-05-12
Tick bit/651 feature request clan role creation modification and granting someone a role should happen through a vote followup

tickBit:
### Brief description

I've been trying to get a hold of this PR.

Fixed clan role vote finalization so successful votes reliably update the player’s role, and invalid voting payloads are handled explicitly instead of silently doing nothing.

Locally `npm run test:ci` may give some errors, but after that `npm run test:ci-retry-failed` gives all green.

I haven't test this in practice, only corrected the 2 errors, that previously persisted.

### Change list

- Fixed clan admin updates so removed admin IDs are not accidentally kept or re-added.
- Ensured clan admin lists only keep players who actually belong to the clan.

- Updated clan role voting finalization to validate missing player_id / role_id data.
- Fixed accepted clan role votes so they reliably update the player’s clanRole_id.
- Removed brittle manual player save/debug logic from clan role vote finalization.

About the checkVotingOnExpire() method.

Now it:

- immediately returns [true, null] if the vote did not go through
- returns a SEReason.REQUIRED error if a player_id or role_id is missing in an approved SET_CLAN_ROLE vote
- updates the player's role via playerBasicService.updateOneById(...)
- removes the test debug and manual findById/save workarounds at the same time

### DIFF
--- a/src/clan/clan.service.ts
+++ b/src/clan/clan.service.ts
@@ -80,7 +80,7 @@ export class ClanService {
 
     if (process.env.NODE_ENV === 'test') {
       clanToCreate.name = `T_${Math.random().toString(36).substring(7, 12)}`;
-      console.log('DEBUG: Test running on DB ->', this.connection.name);
+      // console.log('DEBUG: Test running on DB ->', this.connection.name);
     }
 
     const [clan, clanErrors] = await this.basicService.createOne<Clan, ClanDto>(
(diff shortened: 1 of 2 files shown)


CapoMK25 (commented):
Looks good to me and the main pain point has been to just get the last 2-3 unit tests to pass on the behemoth PR https://github.com/Alt-Org/Altzone-Server/pull/797

My question right now is, what'll happen to the previous 797 PR from here? What should the strategy with that be? Since that PR has the biggest bulk of the implementation for voting right now. I was only left wondering, why a separate PR for this here and not a separate commit with these changes on the 797 PR that's still active with the 2-4 unit test failures still intact?

PR #835, opened 2026-05-20
834 bug sending daily emotion to server returns 500 error

tickBit:
### Brief description

Fixed problem with changing an emotion of a player

### Change list

- fixed one unrelated test, that was mystically broken
- return type of `addEmotion` method fixed in `player.controller.ts`


CapoMK25 (approved):
Approved, should be good to go from my view! Thanks!

PR #837, opened 2026-05-20
836 bug voting data doesnt sometimes return old voting data

tickBit:
### Brief description

Now delay to deleting any vote, should be 1 week.

### Change list

- one somehow changed test
- `clanShop.service.ts`: After finalizing voting, the voting_id isn't delete immediately

### DIFF
--- a/src/clanShop/clanShop.service.ts
+++ b/src/clanShop/clanShop.service.ts
@@ -164,12 +164,6 @@ export class ClanShopService {
 
     await this.votingService.finalizeVoting(voting._id);
 
-    const [, deleteError] = await this.votingService.basicService.deleteOneById(
-      voting._id,
-    );
-
-    if (deleteError) await cancelTransaction(session, deleteError);
-
     await session.commitTransaction();
     await session.endSession();
 
--- a/src/__tests__/clan/ClanService/createOne.test.ts
+++ b/src/__tests__/clan/ClanService/createOne.test.ts
@@ -87,7 +87,7 @@ describe('ClanService.createOne() test suite', () => {
   });
 
   it('Should save Elderly as clan age range', async () => {
-    const elderlyClan = clanCreateBuilder
+    const elderlyClan = ClanBuilderFactory.getBuilder('CreateClanDto')
       .setName('elderlyClan')
       .setAgeRange(AgeRange.ELDERLY)
       .build();


src/clanShop/clanShop.service.ts
```diff
@@ -164,12 +164,6 @@ export class ClanShopService {
 
     await this.votingService.finalizeVoting(voting._id);
 
-    const [, deleteError] = await this.votingService.basicService.deleteOneById(
```
CapoMK25, line 167: Dumb question from me: 

How did this play into returning old voting data? Did this code block cause it? It seems like it to me to the naked eye. The only reason for my question and comment is that this code block is something I've glossed over myself from time to time and never addressed it further. 

This still has the transaction commit to finalize it intact + the `await session.endSession();` as well, so I'm not that terrified about this removal, but I have to make sure just in case.

CapoMK25 (commented):
Otherwise good for me, but I have to wait for output for my comment before I approve, as the code block is something I grapple with myself. 

Good work!

src/clanShop/clanShop.service.ts
tickBit: In other places in the code where the voting is finalized, the voting_id isn't deleted immediately.

For instance in `fleaMarket.service.ts` there is just `await this.votingService.finalizeVoting(voting._id);` without deleting the voting_id. And in `clanRole.service.ts` there is just

```
await this.votingService.finalizeVoting(voting._id);
return [true, null];
```

without deleting the voting_id after finalizing.

When deleting the voting_id is removed in the code block above, the code is similarly implemented for every voting and the voting_id should remain about 1 week.

tickBit: `clanShop.service.ts` has small other bug, too. I will fix it, when this is resolved.

Though, after more research I found some more problems, with the `clanShop` that the `fleaMarket` doesn't have. When those two have similar implementation regarding items both will work.

CapoMK25: What is the functionality of the voting_id and how did it directly relate to sometimes not returning older voting data?

tickBit: As far as I can see, the voting_id is the id of the latest voting. When it's deleted, that particular voting is deleted entirely so, that it can't be seen in voting history.

There is a cron job that cleans up after a week the old votes. At least that is how i interpreted the implementation.

CapoMK25 (approved):
Approved on all fronts the way I see it.

PR #838, opened 2026-05-20
Fix Swagger example values and update DTOs

tickBit:
### Brief description

Correct incorrect Swagger values

### Change list

Corrections in endpoints

- profile
- createProfile
- signin
- auth


CapoMK25 (approved):
Solid work, approved.

PR #843, opened 2026-05-22
player-emotion-cleanup.task.ts removes all emotions objects that are …

valtteri22005:
### Brief description

player-emotion-cleanup.task.ts removes all emotions objects that are more than a week old. This task runs once a day at midnight.

### Change list

- Imported PlayerEmotionCleanupTask, in the player.module.ts -file


CapoMK25 (approved):
Great work, approved.

PR #844, opened 2026-05-22
839 bug in clanshop voting is not calculated with clan majority

tickBit:
### Brief description

When voting for buying an item, it happens with clan majority.
If there's clear majority before the 10 minutes of time, the voting ends.

### Change list

- similar implementation, than in the `fleaMarket.service.ts`
- `clanShop.service.ts`

The following change is important:

```
if (!voting.endedAt) {
  await this.votingService.finalizeVoting(voting._id);
}
return await endTransaction(session, true);

```

These lines:

```
await session.endSession();
```

with

`endTransaction(session, true);`

cause a double commit to DB and runtime failure. `endTransaction` requires `await` before it, too. `await` is added now and those lines, that cause double commit are removed. In other words those two lines and `endTransaction` do the exact same thing. In this implementation, the both would cause a runtime error.


tickBit:
Tested with Postman.

CapoMK25 (approved):
Looks solid, very good work. Approved.

PR #850, opened 2026-05-26
testing a simple CI config first

### LINKED ISSUE #849: Feature request: Jenkins migrated to GitHub Actions

Jenkins currently doesn't accept MFA requests for whatever reason. Nobody currently has access to Jenkins further, it can only run unit tests and the rest of the CI workload just fine, but admin access is currently revoked for an unknown reason. 

It's been discussed that this server should switch back to GitHub Actions instead of Jenkins. Looking at the Actions runners, the latest activity there was from May 2025 exactly one year ago. Thus this is completely possible, just the costs need to be investigated and addressed first before-hand. 

Make a plan of how to pull this off and what else to actually take into account. This touches the repo's deepest settings so we need a clear action plan and steps to achieve this with 0 downtime, if possible. 

## Any additional info?

Currently only Mikhail as the original lead has access to the repo settings, on top of @jpetays and nobody else has access. This should be addressed further as it's been discussed in the past already. The new person with VIP repo access should either be me and/or @tickBit as the leads.

CapoMK25:
### Brief description

This is just a test for GitHub Actions in the future, might fail completely on the first run and might not even work together with Jenkins, since Jenkins is still active.

### Change list

- Added two workflows for the actions runners, thus the 2 added files. `ci.yml` and `coverage.yml`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,78 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - dev
+      - main
+  push:
+    branches:
+      - dev
+      - main
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '24'
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run lint
+        run: npm run lint
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '24'
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build
+        run: npm run build
+
+  test:
+    name: Unit Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '24'
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run tests
+        run: npm run test
+
+      - name: Upload coverage report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-report
+          path: coverage/
+          retention-days: 7
+          if-no-files-found: ignore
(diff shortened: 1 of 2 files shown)


CapoMK25:
Too strict, has to rerun the classic Mongo test isolation bugs instead of failing the run completely. Adjusting.

CapoMK25:
`--runinband` makes the CI loop forever on Actions, needs to be addressed in the coming days by me.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '24'
```
hoan301298, line 24: I think now we are using the lower version of node 20/22. It might be a problem in future. Just choose a safer one. Please check other usage as well. 😄

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run tests (first attempt)
+        id: first_run
+        continue-on-error: true
+        run: npm run test -- --json --outputFile=test-results.json
```
hoan301298, line 71: guarantee coverage exists

```
run: | 
    npm run test -- \ 
    --coverage \ 
    --json \ 
    --outputFile=test-results.json
```

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,33 @@
+name: Coverage
+
+on:
+  push:
+    branches:
+      - dev
+
```
hoan301298, line 7: Prevents old coverage runs from continuing after new commits are pushed
```
concurrency:
  group: coverage-${{ github.ref }}
  cancel-in-progress: true
```

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Run tests with coverage
+        run: npm run test -- --coverage
+
+      - name: Upload coverage to GitHub
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-report-${{ github.sha }}
+          path: coverage/
+          retention-days: 30
```
hoan301298, line 33: The dedicated coverage workflow exists specifically to generate coverage.

If the coverage/ folder is missing there, something is wrong:

tests didn’t run correctly
coverage config broke
Jest stopped generating reports

So failing the workflow is useful:

```
if-no-files-found: error
```

because coverage generation is the primary goal of that workflow.

hoan301298 (approved):
I think the current setup makes sense. Approved! Well done.

In the dedicated `coverage.yml` workflow, using `if-no-files-found: error` also makes sense since generating coverage is the primary goal there.

I added comments, some of them were from AI suggestions. Feel free to discuss and if you think they're no need for this ci workflows, ignore them 👍

.github/workflows/ci.yml
tickBit: I'm currently using Node v25.9.0

CapoMK25:
GH Actions and CI/CD is now in full effect. Closing this PR finally as unneeded from here on out.

PR #856, opened 2026-05-29
819 Addressed the limits of the votes bug

### LINKED ISSUE #819: Bug: Limits of the votes
## What doesn't work?

Currently, there isn't a limit on how many votes a player can send.
Also the player can vote in an expired voting. This does not change a "result", but the server returns a status code 204 and adds the vote to the voting object in the database.

Endpoint: /voting

Method: 

## How can it be reproduced?

Send multiple votes on the same player.

## What should it do instead?

The server should check if the player has voted and is the voting expired.

Suomeksi.
Tällä hetkellä pelaaja voi lähettää useita ääniä, tämä vaikuttaa äänestystulokseen.
Lisäksi pelaaja voi äänestää vanhentunutta äänestystä. Tämä ei vaikuta vanhaan "tulokseen", mutta serveri lähettää vastauskoodin 204 ja lisää äänen voting objektiin.
Serverin tulisi tarkistaa onko pelaaja jo äänestänyt ja onko äänestys vielä auki.

CapoMK25:
Fixes #819 (partial, see the scope note below).

Adds a check to `VotingService.addVote()` that rejects votes on votings
that have already ended. A voting is considered ended if either:

- `endedAt` is set (the voting was finalized early due to threshold
  being reached, or via explicit finalization), or
- `endsOn` is in the past (the 10-minute timer expired but the scheduler
  hasn't run cleanup yet).

In both cases `addVote` now throws `votingExpiredError` (HTTP 400 via
`APIErrorFilter`) instead of silently accepting the vote and returning
204. The vote is no longer added to the voting record.
205. 

### Scope note

Issue #819 originally described two related bugs:

1. A player can vote multiple times on the same voting.
2. A player can vote on an expired voting.

Bug 1 was already fixed in commit c0ecee0a (May 13, 2026) by tickBit:
the existing duplicate-vote check had a missing `.toString()` on the
ObjectId comparison, which made it always evaluate false. That fix is
already in dev. This PR addresses bug 2 only.

### Change list

- `src/voting/voting.service.ts`
  - In `addVote`: add two early-return checks for `endedAt` and `endsOn`,
    placed before the existing duplicate-vote check.
- `src/voting/error/votingExpired.error.ts` (new file)
  - New `votingExpiredError` using SEReason.NOT_ALLOWED, following the
    pattern of `alreadyVotedError`.
- `src/__tests__/voting/VotingService/addVote.test.ts`
  - Updated test "Should not allow voting twice by the same user" to
    pre-seed NO votes so the player's vote doesn't trigger finalization
    (with the new expiration check, a finalized voting correctly rejects
    further votes, which was making the duplicate-vote scenario
    unreachable).
  - Added test "Should not allow voting on an expired voting (endsOn
    in past)" to cover the new behavior.
  - Added test "Should not allow voting on a finalized voting (endedAt
    set)" to cover the new behavior.

### Testing

Unit tests pass (`npm run test -- addVote`). Full suite green except for
the pre-existing flaky tests unrelated to voting (`ClanService/readAll`,
`ItemService/readMany`, `SoulHomeService/deleteOneById`).

### DIFF
--- a/src/fleaMarket/enum/status.enum.ts
+++ b/src/fleaMarket/enum/status.enum.ts
@@ -9,11 +9,11 @@
  *   clan. Lives in the FleaMarketItem collection.
  * - BOOKED: Another clan has reserved the item and is in the process of
  *   voting whether to buy it. Lives in the FleaMarketItem collection.
- * Step 2 next tomorrow, delete this when there. 
+ * Step 2 next tomorrow, delete this when there.
  */
 export enum Status {
   IN_STOCK = 'in_stock',
   AVAILABLE = 'available',
   SHIPPING = 'shipping',
-  BOOKED = 'booked'
+  BOOKED = 'booked',
 }
--- a/src/fleaMarket/enum/status.enum.ts
+++ b/src/fleaMarket/enum/status.enum.ts
@@ -9,7 +9,6 @@
  *   clan. Lives in the FleaMarketItem collection.
  * - BOOKED: Another clan has reserved the item and is in the process of
  *   voting whether to buy it. Lives in the FleaMarketItem collection.
- * Step 2 next tomorrow, delete this when there.
  */
 export enum Status {
   IN_STOCK = 'in_stock',
--- a/src/fleaMarket/enum/status.enum.ts
+++ b/src/fleaMarket/enum/status.enum.ts
@@ -1,10 +1,19 @@
 /**
- * Enum for flea market items.
+ * Lifecycle status of a furniture item.
  *
- * Used in the status field.
+ * - IN_STOCK: Item is in the clan's stock, not currently for sale.
+ *   (Regular Items have this status conceptually but the field may not be persisted.)
+ * - AVAILABLE: Item has been approved for selling via a vote, but not yet
+ *   placed onto a stall. Lives in the FleaMarketItem collection.
+ * - SHIPPING: Item is on the clan's stall, waiting to be bought by another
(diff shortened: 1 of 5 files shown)


src/fleaMarket/enum/status.enum.ts
```diff
@@ -1,9 +1,18 @@
...
- * Used in the status field.
+ * - IN_STOCK: Item is in the clan's stock, not currently for sale.
+ *   (Regular Items have this status conceptually but the field may not be persisted.)
+ * - AVAILABLE: Item has been approved for selling via a vote, but not yet
+ *   placed onto a stall. Lives in the FleaMarketItem collection.
+ * - SHIPPING: Item is on the clan's stall, waiting to be bought by another
+ *   clan. Lives in the FleaMarketItem collection.
+ * - BOOKED: Another clan has reserved the item and is in the process of
+ *   voting whether to buy it. Lives in the FleaMarketItem collection.
+ * Step 2 next tomorrow, delete this when there.
```
tickBit: Probably a forgotten comment. Doesn't cause confusion though, meant to be deleted, I think.

tickBit (approved):
Good work! I think there is one forgotten comment line. I don't see anything unfinished here.

@CapoMK25 you can delete that comment line, if you want, but I approve this.

src/fleaMarket/enum/status.enum.ts
CapoMK25: Will remove later, thanks for the heads-up!

PR #859, opened 2026-06-02
858 feature request add clan name to the playerdata that is received from the server

tickBit:
### Brief description

Clan's name now visible from endpoints with player object.

From Postman:

```
{
    "data": {
        "Player": [
            {
                "_id": "6a043d92aaacc61e8681495f",
                "name": "super 2",
                "points": 0,
                "battlePoints": 0,
                "claimableRewards": [],
                "backpackCapacity": 453,
                "uniqueIdentifier": "1415",
                "above13": true,
                "parentalAuth": null,
                "gameStatistics": {},
                "battleCharacter_ids": [],
                "currentAvatarId": null,
                "profile_id": "6a043d92aaacc61e8681495d",
                "clan_id": "6a019315ee393ee3d344e199",
                "avatar": null,
                "clanRole_id": "6a019315ee393ee3d344e19c",
                "playstyle": "Balanced",
                "carbonFootprint": 0,
                "clanCoinsAccumulated": 0,
                "emotions": [],
                "clanName": "super_clan"
            },
```
			
I tried to keep changes minimal, but I tried to improve the implementation so, that type `any` isn't used. Removing it from 1 endpoint caused small fixes to many places.

### Change list

Clan's name now visible from endpoints

`/player`
`/player/{_id}`

To avoid using type `any` I created a new type `PlayerObject`, too.

Some tests had to be changed a bit too and some other code files. All those changes are related to avoiding using type `any` somehow.

### DIFF
--- a/src/__tests__/player/PlayerService/getPlayerById.test.ts
+++ b/src/__tests__/player/PlayerService/getPlayerById.test.ts
@@ -1,4 +1,5 @@
-import util from 'util';import PlayerBuilderFactory from '../../player/data/playerBuilderFactory';
+import util from 'util';
+import PlayerBuilderFactory from '../../player/data/playerBuilderFactory';
 import PlayerModule from '../../player/modules/player.module';
 import { getNonExisting_id } from '../../test_utils/util/getNonExisting_id';
 import { PlayerService } from '../../../player/player.service';
@@ -10,7 +11,6 @@ import { clearDBRespDefaultFields } from '../../test_utils/util/removeDBDefaultF
 import { Player } from '../../../player/schemas/player.schema';
 
 describe('PlayerService.getPlayerById() test suite', () => {
-  
   let playerService: PlayerService;
   const playerBuilder = PlayerBuilderFactory.getBuilder('CreatePlayerDto');
   const updatePlayerBuilder =
@@ -44,7 +44,7 @@ describe('PlayerService.getPlayerById() test suite', () => {
     const [player, errors] = await playerService.getPlayerById(
       existingPlayer._id,
     );
-    
+
     expect(errors).toBeNull();
     expect(player).toBeDefined();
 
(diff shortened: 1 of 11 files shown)


src/clanShop/clanShop.service.ts
```diff
@@ -25,6 +25,7 @@ import {
 import { InjectConnection } from '@nestjs/mongoose';
 import { IServiceReturn } from '../common/service/basicService/IService';
 import { OnEvent } from '@nestjs/event-emitter';
+import ServiceError from 'src/common/service/basicService/ServiceError';
```
CapoMK25: Remove the src here and use `../common/service/basicService/ServiceError` instead.

src/__tests__/player/PlayerService/getPlayerById.test.ts
```diff
@@ -1,4 +1,4 @@
-import PlayerBuilderFactory from '../../player/data/playerBuilderFactory';
+import util from 'util';import PlayerBuilderFactory from '../../player/data/playerBuilderFactory';
```
CapoMK25: Is this a typo? The import looks very weird.

CapoMK25 (changes requested):
Solid work, only the 2 comments I'd like to see addressed and we're good to go.

src/__tests__/player/PlayerService/getPlayerById.test.ts
tickBit: Probably a typo. I made long day and was getting probably a bit careless : )

I think I needed in one test temporarily `util from 'util'` and forgot the import there somehow.

src/clanShop/clanShop.service.ts
tickBit: I had those more and tried to replace each one of them with a relative path. I made a 10+ hours day, so I was a bit careless : )

tickBit:
I noticed, that all those problematic imports were unused imports. So I just removed them. For some reason, locally only at the second `npm run test:ci-retry-failed` the tests were all green.

tickBit:
Getting tired again: Forgot first to change one import to relative path.

CapoMK25 (approved):
Good to go, good job, approved.

Issue #841, opened 2026-05-21
Feature request: Update the clan stock to include the furnitures that are set to be sold and/or set onto stall


BillTheBeast:
## What do you want to see in the API?

Update the clan stock to include the furnitures that are set to be sold and/or set onto stall (Ie. are currently placed into the fleamarket)

## How do you think this should work?

Currently the clans furniture are split into three groups:
- Stock that contains most of the furniture clan has and can use.
  - These are fetched through the Stock endpoint.
- Furniture that are approved for selling but aren't yet placed onto the stall.
  - These are fetched through the fleamarket endpoint.
    - The problem is that this endpoint contains all furniture in the fleamarket from every clan. This is mostly redundant now since all of the furniture in fleamarket should be fetched with the clans stalls. And also is liable to be clogged due to the number to furniture there would eventually be in this.
- Furniture placed onto stalls.
  - These are fetched with the stalldata.

The second group is currently very problematic due to the things that I noted under it.

By deprecating the second group and moving it under allowing also the furniture that are tagged to be sold (as well as the furniture placed into stalls) also be available through the stock solves a lot of the issues here as that allows clients an easy access to all of the clans furniture regardless where they are.
- Note: The furniture that are placed onto the stall should still also be available through the stall endpoint as this is how other clans would gain access to it. 

The furniture currently has a "status" field that is mainly used to track the status of the furniture when it is approved to be sold. This could be expanded to also include a status for when the furniture is just "normally" in the stock. (InStock - enum). This would make it possible to differentiate if the furniture is being sold or not.

## Any additional info?

The status field values are currently: "available", "shipping" and "booked".
There is lacking documentation on what each of these actually means. This should also be improved.

My personal understanding is that:
"available" = Tagged as to be sold but not on stall yet.
"shipping" = On stall waiting to be bought
"booked" = Another clan has reserved the furniture and are in a process of voting whether to buy the furniture.

But you should double check if this is how it actually works and document them properly.

CapoMK25:
Assigned to myself, I'll start work on this by the latest when my PR is merged, possibly earlier since the risk of touching the exact same files is fairly low, max the `fleamarket.service.ts` if that.

CapoMK25:
@BillTheBeast  Working on #841. Quick clarifications:

For the Stock response shape: should `fleaMarketItems` be a separate field alongside items, or merged into a single flat list with a status field on each?

[...]

PR #863, opened 2026-06-18
841 feature request update the clan stock to include the furnitures that are set to be sold andor set onto stall

CapoMK25:
### Brief description

Should be updated soon by me as soon as I remember what I did a month ago. The unit tests are all green on this finally. 

### Change list

- Will be walked through later. Directly resolves issue #841


tickBit (approved):
Approved 👍 I don't see anything, that might break anything. Looks logical.

src/voting/schemas/voting.schema.ts
```diff
@@ -12,7 +12,11 @@ import { GovernancePayload } from '../type/governancePayload';
 // =================================================================================
 if (typeof mongoose.Model.discriminator === 'function') {
   const originalDiscriminator = mongoose.Model.discriminator;
-  mongoose.Model.discriminator = function (name: string, schema: any, tiebreaker: any) {
+  mongoose.Model.discriminator = function (
+    name: string,
+    schema: any,
```
tickBit, line 17:
```
schema: any,
tiebreaker: any
```
Those `any` types seemed to exist earlier and we do have still those. I would guess, that changing those to a safer type might break the production code in many places and the tests would also require adjusting. So, I think they're ok.

PR #886, opened 2026-07-27
885 bug after reseting testing boxsession the box has still no place for new testers anymore

### LINKED ISSUE #885: Bug: After reseting testing box/session, the Box has still no place for new testers anymore
## What doesn't work?

After a testing box/session has been reset, claiming new tester accounts may still fail with the error:

"Box has no place for new testers anymore. Please, increase the allowed amount of testers to be able to claim new tester accounts"

This happens even though the previous tester accounts were removed during reset. The reset clears tester profiles and players, but the box metadata can still keep the old `testerAccountsClaimed` value, so the box appears full.

Endpoint: `/box/reset`, then tester account claim endpoint

Method: `PUT` for reset, then `POST` for claiming tester account

## How can it be reproduced?

1. Create/configure a testing box with a limited `testersAmount`, for example 10.
2. Start the testing session.
3. Claim tester accounts until `testerAccountsClaimed` reaches `testersAmount`.
4. Reset the testing session using `/box/reset`.
5. Start/configure the session again if needed.
6. Try to claim a new tester account with the shared tester password.
7. The backend returns the "Box has no place for new testers anymore" error.

## What should it do instead?

Resetting a testing session should also reset the tester account claiming state.

After reset, `testerAccountsClaimed` should be set back to `0`, and old claim-related state such as `accountClaimersIds`, `createdClan_ids`, and the previous `testersSharedPassword` should be cleared as appropriate.

A new testing session should allow new tester accounts to be claimed up to the configured `testersAmount`.

The issue is related to box metadata, not necessarily to missing tester profiles or players. The old tester data is removed, but the counter used to enforce tester capacity is not reset, causing the box to behave as if it is still full.

tickBit:
### Brief description

In the testing box method PUT `/box/reset` didn't work as expected: the result was

```
"testersAmount": 2,
"testerAccountsClaimed": 2,
"accountClaimersIds": [],
```				

`"testerAccountsClaimed"` should be 0.

### Change list

- box.service.ts:

All the necessary fields are cleared after PUT `/box/reset`
Also, a new test added.

Some linting changes, too...

Here's the output of Postman testing:

```
"createdClan_ids": [],
"testersAmount": 2,
"testerAccountsClaimed": 0,
"accountClaimersIds": []
```
and also
```
"sessionStage": "Preparing",
"testersSharedPassword": null
```


tickBit:
It seems, that the name of the first commit is accidentally wrong: those were not linting changes, but the actual small changes to fix the bug.

CapoMK25 (approved):
Solid, nice job!

PR #887, opened 2026-07-27
790 feature request update the soulhome api and data structure to match that of what is needed on the clientside

### LINKED ISSUE #790: Feature request: Update the SoulHome API and data-structure to match that of what is needed on the clientside.

Current endpoint (GET soulhome/) has this structure according to swagger.
"SoulHome": {
      "clan_id": "666abc12d1e2f30012bbccdd",
      "_id": "666fabc1d2f0e10012aabbcc",
      "name": "Fortress of Dawn",
      "Room": "string",
      "Clan": "string"
    }
This has the issue that it basically contains no information about the soulhome other than its id and name (which might also be unnecessary, but is also harmless), while the actual important information i.e. what is inside the soulhome itself (room design and furnitures) are completely absent.

The Room field should be changed into an Array that contains data for each of the rooms inside the soulhome.
The data in the Room object should be
{
    "id": string
    "roomposition": int (used to determine the order of the rooms)
    "roomcolour": HexaDecimal
    "wallpaper": string
    "floor": string
    "furnitures": Object-array
    "deactivationTime": time in UTC (only used if the room is being deactivated/removed)
    "soulhome_id": string (mainly just a backup to check that the room actually belongs to the requested soulhome)
}
The number of rooms should match the player count in the clan (30 max) and each time a new player joins the clan a new room is added to the clan.
When a player leaves a clan there is a 24-hour (can be adjsted) grace period before the room is removed/deactivated (more on this below) If another player joins the clan before the grace period end then the room isn't removed and will remain as if nothing has changed.
When the grace period ends there are 2 methods that could be used to handle the removed of the room.
1. As long as the room furniture inside it the room is deactivated, where the room is still visible and interactable but which blocks new furniture being added to the room and only allows removing the furniture inside it back to storage. If the room is empty then it is removed until a new player joins.
2. Room is removed and all of the furniture inside it are moved to the clan storage.

The furniture inside that rooms are the same ones as the ones in storage, forever there are a few extra pieces of data that is important for the furniture inside the soulhome.
"roomid" (tells what room the furniture is in, should be null if not in soulhome)
"location" x,y coordinate array (corresponds to top left slot in furniture, furniture of size 2x2 in slot 0,0 would also cover slots 0,1; 1,0; 1,1)
"furnituresize": x,y ints
"rotation": Enum (front right, left ,back) which way the furniture is rotated in.
"position": Enum (floor, floorByWall, Wall, Ceiling) does this furniture belong on the wall, or is it placed on the ceiling, etc. 
"placedOnId": if the furniture are placed on other furniture (Id of the furniture it is placed on, the furniture they are placed on/in could also have the id of furniture are inside)
"placedOnLocation": x,y int (the spot on the furniture this is placed on) 

Some of these furniture are already in the data structure but are mentioned here due to being important specifically for this functionality.

The rooms have slotgrids in which the furniture can be placed upon, the floor and the roof are the size of 20x8, backwall is the size of 20x6 and side walls are the size of 8x6.

There is currently an endpoint for updating rooms PUT room/ but the endpoint should able to take multiple rooms at the same time (array) if needed and also accept just a singular room if that is all that is required.
The changes in these rooms should then be reflective of the changes inside the storage itself. As if a furniture was removed it is now in storage and vise versa. Also there should eventually be validation that the location the player is trying to place the furniture is actually valid, but this can be done much later as a lot of things are currently still in flux and could be changed.

## What is the priorite for which of the sections here?

1:
- Update the soulhome to include the rooms.
- Update the rooms to properly contain the furniture (if they currently aren't there.)
- Update the furniture to have the "roomid, location, position, rotation" fields (position doesn't need to be sent necessarily as this value would be static anyway, but is important for validation on serverside)
- Update the the room update endpoint to take the soulhome status properly.

2.
- Make sure the room count scales with playercount.
- Implement room adding
- Implement the room removal and the grace period.

3.
- Basic Furniture location validation. (Check is the main slot itself already has an furniture and that you aren't trying to place something outside of the valid slots)
- Calculate the value of all the furniture in the soulhome and store it into Clandata (this is needed in a lot of places later)

4.
- Add room colour, wallpaper and floor styles into the room data.

5.
- Implement the data to place furniture on top of each other. (placedOnId, placedOnLocation)

6.
- Further Furniture location validation.

Iir0R:
### Brief description
Updated code to meet requirements of Issue.

### Change list

- Updated soulHome to include rooms and furniture
- Updating room data and furniture implemented
- Updated Items to have rotation, position, placedOn_id, placedOnLocation
- Rooms updated with roomColour, wallpaper, floor
- SoulHome Room creation, activation and deactivation implemented
- Timed inactive Room removal implemented and sending furniture to Stock
- Added furnitureTotalValue to Clan

NOTE: Furniture location isn't validated yet as not all items have sizes.
NOTE: Currently, room Put can't accept array, only single object. Authorize can't get id from array, will be changed in separate update.

Work on this task was started some time ago, so I'm not sure how it will behave with new changes. Creating a PR to see it'll work.


src/clanInventory/room/dto/room.dto.ts
```diff
@@ -1,6 +1,10 @@
-import { Expose } from 'class-transformer';
+import { Expose, Type } from 'class-transformer';
 import AddType from '../../../common/base/decorator/AddType.decorator';
 import { ExtractField } from '../../../common/decorator/response/ExtractField';
+import { ApiProperty } from '@nestjs/swagger';
+import { IsEnum, IsHexadecimal } from 'class-validator';
+import { ItemSummaryDto } from 'src/clanInventory/item/dto/itemSummary.dto';
```
CapoMK25: Swap this for ../ or equivalent prefix instead of src.

src/clanInventory/room/dto/updateRoom.dto.ts
```diff
@@ -1,11 +1,13 @@
...
+  IsHexadecimal,
   IsMongoId,
-  IsNumber,
   IsOptional,
   IsString,
 } from 'class-validator';
 import AddType from '../../../common/base/decorator/AddType.decorator';
+import { Type } from 'class-transformer';
+import { ApiProperty } from '@nestjs/swagger';
+import { ItemSummaryDto } from 'src/clanInventory/item/dto/itemSummary.dto';
```
CapoMK25: Same thing here.

src/clan/utils/defaultValues/items.ts
```diff
@@ -1,6 +1,8 @@
+import { ItemRotation } from 'src/clanInventory/item/enum/itemRotation.enum';
```
CapoMK25: Same thing here.

src/clan/utils/defaultValues/items.ts
```diff
@@ -1,6 +1,8 @@
...
 import { itemProperties } from '../../../clanInventory/item/const/itemProperties';
 import { CreateItemDto } from '../../../clanInventory/item/dto/createItem.dto';
 import { ItemName } from '../../../clanInventory/item/enum/itemName.enum';
+import { ItemPosition } from 'src/clanInventory/item/enum/itemPosition.enum';
```
CapoMK25: Same thing here.

CapoMK25 (approved):
Solid, approved.

tickBit:
Looks good and hard work. I will take closer look by Monday at the latest, if PR isn't merged before that.
Patch coverage is really high 👍

CapoMK25 (approved):
Good work, approved.

PR #891, opened 2026-08-05
890 bug get itemsteal returns 500 internal server error if no request body is provided

### LINKED ISSUE #890: Bug: GET /item/steal returns 500 internal server error, if no request body is provided
## What doesn't work?

`GET /item/steal` returns `500 Internal Server Error` when the request does not include `steal_token` or the request body is missing/undefined.

The likely bug is in `StealTokenGuard`:

```ts
const stealToken = request.query.steal_token ?? request.body.steal_token;
```

For a GET request, `request.body` may be `undefined`. In that case, accessing `request.body.steal_token` throws a `TypeError` before the guard can return the intended authorization error.

Expected result: the endpoint should return `403 Forbidden` with the existing `steal_token is not provided` API error.

Actual result: in some environments, the endpoint returns `500 Internal Server Error`.

Endpoint: `/item/steal`

Method: `GET`

## How can it be reproduced?

1. Send a `GET` request to `/item/steal`.
2. Do not include `steal_token` in the query string.
3. Make sure the request has no body.
4. Observe that the server may return `500 Internal Server Error`.

Likely fix area: `src/clanInventory/item/guards/StealToken.guard.ts`, where `request.body` should be accessed safely.

constf03:
### Brief description

This PR should fix a bug with endpoint GET /item/steal where it returns internal server error if no request body is provided instead of throwing unauthorized API Error like was intended to do.

### Change list

- updated `src/clanInventory/item/guards/StealToken.guard.ts`
  - check that `request.body.steal_token` or `request.query.steal_token` is not undefined before accessing `stealToken` (avoid TypeError that caused internal server error) 

- updated `src/__tests__/clanInventory/item/guards/AuthGuard/canActivate.test.ts`
  - added a new test that checks if Unauthorized APIError is thrown if no request body is provided

### DIFF
--- a/src/__tests__/clanInventory/item/guards/AuthGuard/canActivate.test.ts
+++ b/src/__tests__/clanInventory/item/guards/AuthGuard/canActivate.test.ts
@@ -64,6 +64,21 @@ describe('StealTokenGuard.canActivate() test suite', () => {
     await expect(throwsNotAuthorized).rejects.toThrow(expectedException);
   });
 
+  it('Should throw NOT_AUTHORIZED APIError if the request body is not provided', async () => {
+    const context = contextBuilder
+      .setHttpRequest({ query: undefined, body: undefined })
+      .build();
+
+    const throwsNotAuthorized = async () =>
+      await stealGuard.canActivate(context);
+
+    const expectedException = new APIError({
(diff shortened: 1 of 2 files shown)


tickBit (approved):
Clean and good job! Approved.

PR #898, opened 2026-08-07
894 bug battlestart returns 500 when custom matchid is not a mongo objectid

### LINKED ISSUE #894: Bug: /battle/start returns 500 when custom matchId is not a Mongo ObjectId
## What doesn't work?

`POST /gameData/battle/start` accepts `matchId` as any string in the DTO, but `GameDataService.registerBattle()` uses that value as the Mongo document `_id`.

If `matchId` is something like `"test-steal-001"`, Mongoose validation fails because `_id` expects an ObjectId-compatible value. The API returns a 500 server error instead of a clear client error.

Endpoint: `/gameData/battle/start`

Method: `POST`

## How can it be reproduced?

1. Send a request with body:

```json
{
  "gameType": "matchmaking",
  "team1": ["<valid_winner_player_id>"],
  "team2": ["<valid_loser_player_id>"],
  "matchId": "test-steal-001"
}
```

2. Observe that the server returns 500.
3. The terminal shows a Mongoose `Game validation failed` error from `GameDataService.registerBattle`.

## What should it do instead?

The endpoint should either:

- reject non-ObjectId `matchId` values with a 400 response, or
- support custom string match IDs by storing them in a separate `matchId` field instead of `_id`.

The DTO documentation currently suggests custom match IDs like `"match_12345"`, but that does not work with the current implementation.

tickBit:
### Brief description

There used to be allowed in the `startBattle.dto.ts` `@IsString` custom matchId, but using custom string caused 500 internal server error. To accept custom string without 500 internal server error requires changes to database schema.

Now `matchId` field must be MongoDB style string. It can be custom string with if it's also a 24 byte MongoDB string at the same time.

I `git stash`ed the prettier/linting changes of other files.

Issue #894 

### Change list

- startBattle.dto.ts (require MongoID for matchId)
- added new test file for startBattle.dto.ts

### DIFF
--- a/src/__tests__/gameData/dto/startBattleDto.test.ts
+++ b/src/__tests__/gameData/dto/startBattleDto.test.ts
@@ -1,0 +1,47 @@
+import 'reflect-metadata';
+import { validate } from 'class-validator';
+import { StartBattleDto } from '../../../gameData/dto/startBattle.dto';
+import { GameType } from '../../../gameData/enum/gameType.enum';
+
+describe('StartBattleDto test suite', () => {
+  const player1Id = '665af23e5e982f0013aa1111';
+  const player2Id = '665af23e5e982f0013aa2222';
+  const matchId = '665af23e5e982f0013aa9999';
+
+  function createDto(overrides: Partial<StartBattleDto> = {}) {
+    const dto = new StartBattleDto();
+    dto.gameType = GameType.MATCHMAKING;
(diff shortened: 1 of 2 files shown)


CapoMK25 (approved):
Solid, approved. Be wary in the future with ObjectID vs string mismatches just in case, this PR/issue combo won't likely introduce them, but they're a common pain point in unit testing on bigger issues regardless.

PR #900, opened 2026-08-10
895 bug battle match id handling is inconsistent between start result and conflict resolution

### LINKED ISSUE #895: Bug: Battle match ID handling is inconsistent between start, result, and conflict resolution
## What doesn't work?

Battle match IDs are handled inconsistently in different parts of the service.

`registerBattle()` stores the match identifier as `_id`.

`handleBattleResult()` looks up the battle with `findById(dto.matchId)`.

`resolveConflict()` looks up the battle with `findOne({ matchId })`.

Since `matchId` does not appear to be defined as a separate field in the `Game` schema, `resolveConflict()` may fail to find battles that were created successfully.

Endpoint: `/gameData/battle/start`, `/gameData/battle/result`

Method: `POST`, `PUT`

## How can it be reproduced?

1. Start a battle successfully through `/gameData/battle/start`.
2. Submit conflicting battle results so the battle enters `PROCESSING`.
3. Wait for the conflict resolution timer to run.
4. Observe that `resolveConflict()` searches by `{ matchId }` instead of `_id`.

## What should it do instead?

The battle identifier should be modeled and queried consistently.

Either:

- use Mongo `_id` everywhere, including conflict resolution, or
- introduce a real `matchId` field in the schema and use it consistently across start, result submission, and conflict resolution.

This should probably be decided as a data model/API contract issue before changing the individual queries.

tickBit:
### Brief description

Fixes issue #895 : now MongoID is required and used consistently.

### Change list

- `battleResut.dto.ts` and `submitBattleResult.dto.ts`: Require matchId to be MongoID
- `resolveConflict.ts` look for `_id` in the database instead of `matchId`
- added 2 new test files to cover changes and battle life cycle (resolve conflict)

### DIFF
--- a/src/__tests__/gameData/dto/battleMatchIdDto.test.ts
+++ b/src/__tests__/gameData/dto/battleMatchIdDto.test.ts
@@ -1,0 +1,72 @@
+import 'reflect-metadata';
+import { validate } from 'class-validator';
+import { BattleResultDto } from '../../../gameData/dto/battleResult.dto';
+import { SubmitResultDto } from '../../../gameData/dto/submitResult.dto';
+import { RequestType } from '../../../gameData/enum/requestType.enum';
+
+describe('Battle matchId DTO validation test suite', () => {
+  const matchId = '665af23e5e982f0013aa9999';
+  const player1Id = '665af23e5e982f0013aa1111';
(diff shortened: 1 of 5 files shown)


CapoMK25 (approved):
Solid in my eyes, approved.

PR #902, opened 2026-08-10
901 bug in matchmaking custom match does not require roomid and small other bugs in matchmaking

### LINKED ISSUE #901: Bug: In matchmaking CUSTOM match does not require roomId and small other bugs in matchmaking
## What doesn't work?

There are a few issues in the matchmaking-only flow.

First, joining a `CUSTOM` matchmaking invite does not require the client to provide the `roomId`, even though the DTO documentation describes it as an extra guard for custom joins.

`JoinMatchmakingInviteDto.roomId` is optional, and `validateJoinInvite()` only checks the room ID if it is present. This means a player can join a custom invite without proving they are joining through the expected custom room/lobby.

Second, `finishMatch()` can leak leaderboard update failures as thrown errors. `updateLeaderboardsForFinishedMatch()` throws returned `ServiceError[]` values instead of returning them through the normal `IServiceReturn` flow. If a player or clan leaderboard update fails, the endpoint may return an internal server error instead of a controlled service error response.

There is also a smaller documentation/comment mismatch: `saveFinishedMatch()` says it removes per-player active-match pointers, but the implementation only extends/expires them with a TTL.

Endpoint: `/matchmaking/invites/:inviteId/join`

Method: `POST`

Endpoint: `/matchmaking/matches/:matchId/finish`

Method: `POST`

## How can it be reproduced?

1. Create a `CUSTOM` matchmaking invite with a valid `roomId`.
2. Try joining the invite without sending `roomId` in the request body.
3. Observe that the join can pass because `validateJoinInvite()` only compares `roomId` when `body.roomId` exists.
4. For the finish issue, create or mock an active match.
5. Make `playerService.updatePlayerById()` or `clanService.basicService.updateOneById()` return service errors.
6. Call `POST /matchmaking/matches/:matchId/finish`.
7. Observe that the service throws the errors instead of returning them through the normal service return path.

## What should it do instead?

For `CUSTOM` joins, the service should require `roomId` and reject missing or mismatching values with a clear validation/service error.

For match finishing, leaderboard update failures should be returned as controlled service errors instead of being thrown and potentially becoming 500 responses.

The `saveFinishedMatch()` comment should also be updated to match the implementation, or the implementation should actually remove the per-player active-match pointers if that is the intended behavior.

This appears to be separate from the GameData `matchId` issue. Matchmaking itself generates MongoId-compatible match IDs, so the current `@IsMongoId()` direction is compatible with matchmaking match IDs.

tickBit:
### Brief description

Fixes issue #901 .

In the `joinMatchmakingInvite.dto.ts` `roomId` is required and related error for missing `roomId` is handled.
Also in `updateLeaderboardsForFinishedMatch()` (`matchmaking.service.ts`) errors are returned like in the reset of code (IServerReturn-style).

Documentation improvements.

### Change list

- `matchmaking.service.ts`
- `joinMatchmakingInvite.dto.ts`
- test code additions

### DIFF
--- a/doc/matchmaking-mqtt-notifications.md
+++ b/doc/matchmaking-mqtt-notifications.md
@@ -5,8 +5,8 @@ The implementation lives in `src/matchmaking/matchmaking.notifier.ts`, and the
 state changes that trigger the events are handled by
 `src/matchmaking/matchmaking.service.ts`.
 
-Matchmaking topics currently do not start with `/`. Frontend subscriptions must
-use the exact topic strings shown below.
+Matchmaking topics currently start with `/`. Frontend subscriptions must use
+the exact topic strings shown below.
 
 ## Subscribe Topics
 
--- a/src/__tests__/matchmaking/MatchmakingService/flow.test.ts
+++ b/src/__tests__/matchmaking/MatchmakingService/flow.test.ts
@@ -4,6 +4,7 @@ import { MatchType } from '../../../matchmaking/enum/matchType.enum';
 import { TeamSide } from '../../../matchmaking/enum/teamSide.enum';
 import { MatchmakingService } from '../../../matchmaking/matchmaking.service';
 import { ActiveMatch } from '../../../matchmaking/type/activeMatch.type';
+import { SEReason } from '../../../common/service/basicService/SEReason';
 
 class InMemoryRedisService {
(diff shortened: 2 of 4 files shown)


CapoMK25 (approved):
Nice work, approved!

PR #906, opened 2026-08-11
896 bug battle start validation and authorization are incomplete

### LINKED ISSUE #896: Bug: Battle start validation and authorization are incomplete
## What doesn't work?

`POST /gameData/battle/start` validates that `team1` and `team2` contain MongoId-shaped strings, but it does not appear to verify that the referenced players actually exist.

It also does not appear to verify that the requester is allowed to start a battle with the provided players/teams.

Some invalid states are only caught later by Mongoose schema validation, for example the same player appearing in both teams, which can result in an internal server error style response instead of a clean API validation response.

Endpoint: `/gameData/battle/start`

Method: `POST`

## How can it be reproduced?

1. Send a request with syntactically valid but non-existing player IDs in `team1` and/or `team2`.
2. Observe that the request can pass DTO validation.
3. Send a request where the same valid player ID appears in both `team1` and `team2`.
4. Observe that the duplicate-player check happens at Mongoose validation time.

## What should it do instead?

The endpoint should validate the battle start request at the API/service level before saving.

It should check at least:

- referenced players exist
- the same player is not in both teams
- the requester is allowed to create the battle for those players/teams
- validation failures return clear 400/403 responses instead of uncaught server errors

There is also a small implementation/documentation mismatch around `gameType`: `registerBattle()` sets `gameType: 'BATTLE'`, but then spreads `...dto` afterwards, so the client-provided `gameType` overwrites it.

tickBit:
### Brief description

Fixes issue #896 

Fixes battle start/result validation so battle documents are created with the correct MongoId-backed match identity, matchmaking battles are tied to active matchmaking matches, and only valid participants can start or submit battle results.

### Change list

- Require MongoId-compatible `matchId` values in battle start/result DTOs when provided.
- Store battle `matchId` as the Game document `_id` and use `_id` for battle result/conflict lookups.
- Require `matchId` for `gameType: matchmaking`; keep server-generated ids for casual/custom battles.
- Validate matchmaking battle starts against active matchmaking matches.
- Prevent battle creation when requester/teams/players are invalid.
- Persist only explicit battle start fields instead of spreading the request DTO.
- Reject battle result submissions from non-participants, duplicate submitters, or completed battles.
- Add service and DTO tests for the battle lifecycle validation paths.

### DIFF
--- a/src/__tests__/gameData/gameDataService/battleLifecycle.test.ts
+++ b/src/__tests__/gameData/gameDataService/battleLifecycle.test.ts
@@ -11,10 +11,21 @@ describe('GameDataService battle lifecycle test suite', () => {
   const gameDataModel = GameDataModule.getGameModel();
 
   let gameDataService: GameDataService;
+  let matchmakingService: {
+    validateBattleStart: jest.Mock;
+  };
 
   beforeEach(async () => {
     await gameDataModel.deleteMany({});
     gameDataService = await GameDataModule.getGameDataService();
+    matchmakingService = {
+      validateBattleStart: jest.fn(async () => null),
+    };
+    (
+      gameDataService as unknown as {
+        matchmakingService: typeof matchmakingService;
+      }
+    ).matchmakingService = matchmakingService;
     jest
       .spyOn(gameDataService.playerService, 'getPlayerById')
       .mockImplementation(async (playerId: string) => [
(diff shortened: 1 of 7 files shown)


tickBit:
There’s a matchmaking test on Friday. This PR may be critical for matchmaking to work properly.

CapoMK25 (approved):
Solid, appoved!

PR #914, opened 2026-08-14
913 feature request refactor matchmaking functionality endpoint and mqtt messages

### LINKED ISSUE #913: Feature request: Refactor matchmaking functionality, endpoint and MQTT messages.

A noticed that a lot of the endpoints and notifications used weird nomenclature as the names imply that are about inviting players into room while that actual functionality is just about creating these room and has basically nothing in relation to actual invites in them.

Thus these endpoints and notifications should be renamed to better match their actual purpose.

Also actual functionality should be added to handle actual invites to help players making into the the rooms are desired.
These invites should be depending on the situation be sent either to specific players or to every one in the clan. Note that the clan-wide invites should be handled a little differently from the player specific ones in client-side thus they need to be marked slightly differently. ( 'INVITE_RECEIVED' or 'CLAN_INVITE_RECEIVED' should work for the notification types)

Also the matchmaking probably shouldn't progress automatically, even if it should be possible for the server to do so, due to the clients not yet being in the correct state. This is especially true for the match start, since this requires the players to join the Photon rooms in client-side to actually start the match, hence requiring the for the players to actually be ready before marking the match as started.

## Changes to REST endpoints

- Rename all of the /matchmaking/invites endpoint to /matchmaking/rooms to clarify that they are about the rooms themselves and not the actual invites.
  - Room creation body should include an optional field for automatic invite settings (if CLAN the invite is sent to everybody in the clan, if PLAYER and id is given then an invite is automatically sent to the specified player on room creation.) 

- Add a new POST /matchmaking/invites/{playerId} endpoint that can be used to send invites to other specified players
- Also add POST /matchmaking/invites/clan to send the invite to everybody in clan.

- Add POST /matchmaking/room/start to start the matchmaking. This should be sent by the room owner and is mainly used to make sure that the all of the players are ready to start matchmaking properly.

- Add POST /matchmaking/matches/{matchId}/start to start the match. This should be sent by the players that they have successfully joined the Photon Room used to start the Quantum-simulation AKA the actual clientside battle.

## Changes to the MQTT messages

/matchmaking/invites/player/{playerId} -> /matchmaking/rooms/player/{playerId}
- The type of the message should be changed to ROOM_UPDATED instead of INVITE_UPDATED
- This receives information about all of the available rooms that are available to the player and their updates.
- However the current payload includes multiple id field for the room/match when only one is likely required.

Add a new /matchmaking/invites/player/{playerId} topic
- This is solely used for when the player is invited to a specific room

{
  type: 'INVITE_RECEIVED' or 'CLAN_INVITE_RECEIVED', (Former for player specific invites and latter for whole clan invites.)
  payload: 
  {
    id: string,
    matchType: 'RANDOM' | 'CLAN' | 'CUSTOM',
    status: 'OPEN' | 'READY' | 'QUEUED' | 'MATCHED' | 'CANCELLED',
    ownerPlayerId: string,
    teamSize: 1 | 2,
    allowBots: boolean,
    sentAt: string,
  }
}
This is likely the payload that is required for the invite: The room id, MATCHTYPE, ownerPlayerId (presumably the invitee, if other players can also send invites the have these as a separate field), teamSize, are bots allowed and the timestamp of when the invite was sent. Status can also be included just in case.

Other notifications are mostly fine. Just make sure that the MATCH_STARTED is sent as a response to the "Ready" calls from the clients.

tickBit:
### Brief description

Updates matchmaking room and invite handling so rooms use clearer `/matchmaking/rooms` naming, explicit invites have their own endpoints and MQTT topic, matchmaking starts only when the room owner starts a ready room, and clientside battle start is confirmed by match participants after joining the Photon Room.

Related issue #913 

### Change list

- Renamed matchmaking room HTTP endpoints from invite wording to `/matchmaking/rooms`.
- Added `POST /matchmaking/invites/{playerId}` for sending a room invite to a specific player.
- Added `POST /matchmaking/invites/clan` for sending room invites to available clan members.
- Added room-owner matchmaking start flow with `POST /matchmaking/rooms/{roomId}/start`.
- Added `POST /matchmaking/matches/{matchId}/start` for Photon Room readiness confirmation.
- Changed `MATCH_STARTED` publishing so it is sent only after all real match players have confirmed readiness.
- Added automatic invite settings to room creation for player-specific and clan-wide invites.
- Updated matchmaking MQTT room updates to publish `ROOM_UPDATED` on `/matchmaking/rooms/player/{playerId}`.
- Added explicit invite MQTT notifications on `/matchmaking/invites/player/{playerId}` with `INVITE_RECEIVED` and `CLAN_INVITE_RECEIVED`.
- Added compact MQTT room update payload without duplicate room or match id fields.
- Added and updated tests for room start, match start readiness, invite notifications, and MQTT notification contracts.
- Added separate matchmaking API and MQTT documentation.


tickBit:
In the original issue was asked to send "MATCH_STARTED" after ready calls. To avoid server-side confusion "BATTLE_STARTED" is sent.

BillTheBeast:
I just realised one thing that should be checked. 
Does the invite sending block sending another invite to the same player? 
If it doesn't then all is basically good, but there maybe could be an endpoint (and subsequent notification) for rejecting the invite.
POST /matchmaking/invites/{matchId}/reject and the notification (INVITE_REJECTED, with the playerId)  could be sent through the  /matchmaking/matches/player/{playerId} MQTT line.

tickBit:
> I just realised one thing that should be checked.
> Does the invite sending block sending another invite to the same player?
> If it doesn't then all is basically good, but there maybe could be an endpoint (and subsequent notification) for rejecting the invite.
> POST /matchmaking/invites/{matchId}/reject and the notification (INVITE_REJECTED, with the playerId) could be sent through the /matchmaking/matches/player/{playerId} MQTT line.

The current invite sending does **not** block sending another invite to the same player. Invites are notification-style messages for an existing matchmaking room, and the backend does not currently persist a per-recipient invite state that would prevent duplicates.

So the current implementation should be fine for the “send another invite to the same player” part.

A reject endpoint is not implemented yet. If we want explicit rejection support, it would likely need a separate follow-up change, for example: `POST /matchmaking/invites/{matchId}/reject`

BillTheBeast:
> A reject endpoint is not implemented yet. If we want explicit rejection support, it would likely need a separate follow-up change, for example: `POST /matchmaking/invites/{matchId}/reject`

It's not important right now but it could be something to think about in the future.

BillTheBeast:
Also I just realised another thing, is there a way to change some of the settings of the game room after the room has already been created. (There might be an interest to change the bot settings or the game mode after the fact)

tickBit:
> Also I just realised another thing, is there a way to change some of the settings of the game room after the room has already been created. (There might be an interest to change the bot settings or the game mode after the fact)

Currently, no. The implemented flow does not include an endpoint for updating room settings after the room has been created.

After creation, the supported room actions are basically join, start matchmaking, cancel, and send invites.

If we want editable room settings, that should probably be a separate follow-up endpoint, for example:
`PATCH /matchmaking/rooms/{roomId}`

BillTheBeast:
We can think about that later. But it should be done at some point since the custom matches in particular are likely to change after creation (especially the bot status right now, maybe Locked/Password settings and if separate gamemodes are implemented later it would likely be smart to be able to change these)

BillTheBeast (approved):
Quickly looking seems to be correct. Though also through testing we'll know if it functions properly.

tickBit:
I updated my branch with the base branch. It seems, that someone from the server team still have to approve this.

tickBit:
I'm still updating this a bit.

tickBit:
**Update**

The matchmaking MQTT payloads have been reduced so real player data is now sent only in the compact UI shape:
```
{
  playerId: string,
  name: string,
  avatar: AvatarDto | null
}
```
This applies to:
- ROOM_UPDATED.payload.players[]
- INVITE_RECEIVED.payload.ownerPlayer
- INVITE_RECEIVED.payload.senderPlayer
- CLAN_INVITE_RECEIVED.payload.ownerPlayer
- CLAN_INVITE_RECEIVED.payload.senderPlayer
- MATCH_FOUND.payload.teams[].players[]
- MATCH_STARTED.payload.teams[].players[]
- MATCH_FINISHED.payload.teams[].players[]

For ROOM_UPDATED, `ownerPlayerId` is still included as a plain string ID: `ownerPlayerId: string`

This is intentionally just the owner’s player ID, not a full player object. The full list of room players is available in `players[]` using the compact `{ playerId, name, avatar }` shape. Bots remain separate in `bots[]` as `{ botId, displayName, isBot: true }`.

src/matchmaking/matchmaking.service.ts
```diff
@@ -1293,4 +1860,59 @@ export class MatchmakingService {
...
+  }
+
+  private toMqttPlayerDto(
+    playerId: string,
+    player: { name?: string; avatar?: unknown } | null,
+  ): MatchmakingMqttPlayerDto {
+    return {
+      playerId,
+      name: player?.name ?? '',
+      avatar: (player?.avatar as AvatarDto | undefined) ?? null,
```
CapoMK25: Should these be addressed as something else besides `undefined` or `unknown`? Since the implementation is underway, it'd be good to address these by removing the cast by fixing the Player type so avatar is already `AvatarDto | undefined | null` instead of unknown at least?

CapoMK25 (commented):
The comment could be addressed, prolly shoud be, otherwise solid work all around, great job!

src/matchmaking/matchmaking.service.ts
tickBit: Good point 👍 ! Player's avatar exists or is null. Just in case we could also keep the undefined case. I'll make that change.

tickBit: In fact I'll change it into form `avatar: player?.avatar ?? null`. This takes the `undefined` case in count. This is clean for the frontend: Avatar either exists or is `null` in this payload.

Edit:
The entire function is then
```
private toMqttPlayerDto(
    playerId: string,
    player: { name?: string; avatar?: AvatarDto | null } | null,
  ): MatchmakingMqttPlayerDto {
    return {
    playerId,
    name: player?.name ?? '',
    avatar: player?.avatar ?? null,
  };
}
```

tickBit:
I consider this PR ready, if no further improvements are required.

CapoMK25 (approved):
Nice one, approved! Merging.

PR #916, opened 2026-08-14
912 feature request delete chat messages after 3 days

### LINKED ISSUE #912: Feature request: Delete chat messages after 3 days

Automatically delete chat messages from the server once they are older than 3 days.

Chat messages should have a retention period of 3 days based on their creation timestamp. After that, they should be removed automatically without requiring a manual API call.

A good implementation could use a MongoDB TTL index on the chat message `createdAt` field, or an equivalent scheduled cleanup job if TTL indexing is not suitable for this project. The cleanup should apply consistently to all chat message types, including global, clan, and private messages.

## Any additional info?

Currently the project appears to support deleting individual chat messages manually in `TESTING_SESSION`, but there does not seem to be automatic cleanup for old chat messages.

constf03:
### Brief description
This PR makes new chat messages get deleted from the database automatically after 3 days.
We discussed that existing old documents in the chatMessages table in the production database could be removed with a script which in this case could be just running a query  in the production db like `db.chatMessages.deletemany({})` or something like this. This purge should be done after the next dev to main branch merge assumed this PR is merged into dev.

### Change list

- Updated `src/chat/schema/chatMessage.schema.ts`
  - Added TTL index for `createdAt` field (expire after 3 days)

### DIFF
--- a/src/chat/schema/chatMessage.schema.ts
+++ b/src/chat/schema/chatMessage.schema.ts
@@ -7,6 +7,8 @@ import { Reaction } from './reaction.schema';
 import { Feeling } from '../enum/feeling.enum';
 import { ObjectId } from 'mongodb';
 
+const CHATMESSAGE_TTL = 60 * 60 * 24 * 3;
+
 export type ChatMessageDocument = HydratedDocument<ChatMessage>;
 
 @Schema({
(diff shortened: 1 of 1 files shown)


tickBit (approved):
I double checked, that we have setting autoIndex: true for MongoDB.

Good job 👍 ! Approved.

PR #917, opened 2026-08-15
905 bug put room does not reject cross clan room updates

### LINKED ISSUE #905: Bug: PUT /room does not reject cross-clan room updates
## What doesn't work?

`PUT /room` does not properly reject cross-clan room updates. A user from another clan can still pass the authorization check for a room that belongs to a different clan.

In the current retest, the same scenario returns `500 Internal Server Error` instead of a controlled authorization/not-found response. The `500` appears to be a separate regression caused by the update flow assuming `furniture` is always present in the request body.

Endpoint: `/room`

Method: `PUT`

## How can it be reproduced?

1. Create or use two clans, Clan A and Clan B.
2. Create/use a user who belongs to Clan A.
3. Get a room `_id` that belongs to Clan B’s SoulHome.
4. Send `PUT /room` as the Clan A user using Clan B’s room `_id`.
5. Use a payload such as a room color/floor/wallpaper update, especially without `furniture`.
6. Observe that the request returns `500 Internal Server Error` in the current retest.

## What should it do instead?

The request should be rejected with a controlled client error. Based on the endpoint documentation, the expected response should be `404 NOT_FOUND` when the room does not belong to the logged-in user’s clan.

A `403 NOT_AUTHORIZED` response would also be reasonable if the authorization layer is intended to handle this case, but it should not return `204` or `500`.

Current room authorization appears to check only whether the room exists by `_id`, not whether the room belongs to the logged-in user’s clan/SoulHome. Also, `updateSoulHomeRooms()` assumes `room.furniture` exists and calls `.map()` on it, which may explain the current `500` when the payload omits `furniture`.

constf03:
### Brief description

This PR should fix two bugs:

1. Cross clan room updates (PUT /room)
2. Endpoint PUT /room returning internal server error 500 if furniture is not in the payload.

### Change list

- updated `roomRules.ts`
  - throw 403 ForbiddenException if SoulHome.clan_id and logged in user's clan ID do not match
- updated `room.service.ts`
  - check that `furniture` exists otherwise fallback to empty list
- updated `updateSoulHomeRooms.test.ts`
  - add new test that checks Room values is updated successfully if furniture is excluded from the payload
- added new test suite `roomRules.test.ts`

### DIFF
--- a/src/__tests__/authorization/rule/roomRules.test.ts
+++ b/src/__tests__/authorization/rule/roomRules.test.ts
@@ -1,0 +1,185 @@
+import { ForbiddenException, NotFoundException } from '@nestjs/common';
+import { roomRules } from '../../../authorization/rule/roomRules';
+import { Action } from '../../../authorization/enum/action.enum';
+import { RequestHelperService } from '../../../requestHelper/requestHelper.service';
+import { RoomDto } from '../../../clanInventory/room/dto/room.dto';
+import { UpdateRoomDto } from '../../../clanInventory/room/dto/updateRoom.dto';
+import { SoulHomeDto } from '../../../clanInventory/soulhome/dto/soulhome.dto';
+import { Player } from '../../../player/schemas/player.schema';
+import AuthBuilderFactory from '../../auth/data/authBuilderFactory';
+
+describe('roomRules() test suite', () => {
+  const userBuilder = AuthBuilderFactory.getBuilder('User');
+
+  const roomId = 'room-id';
+  const soulHomeId = 'soulhome-id';
+  const playerId = 'player-id';
+  const clanId = 'clan-a-id';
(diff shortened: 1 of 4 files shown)


constf03:
code coverage % is insufficient so I need to create a unit test for roomRules.ts since currently none of the *Rules.ts files have unit tests. Possible future issue idea to create tests for other rules classes too?

CapoMK25 (approved):
Solid work in my eyes, props for the `ForbiddenException` as well, that is nicher and should work better for this use case. Approved!

CapoMK25:
You can try getting past the codecov/project runner by pushing an empty commit, doesn't always work, but is the lowest hanging fruit for expected-style cases.

tickBit (approved):
Nice solution 👍 ! My approval, too.

PR #923, opened 2026-08-19
842 friend status and requests

tickBit:
### Brief description

Implements MQTT events for friendship status changes and requests.

### Change List

- Removed friendship MQTT notification logic from `FriendshipService`.
- Changed `FriendshipService.addFriend()` to return the created friendship document instead of sending MQTT directly.
- Moved new friend request notification orchestration to `FriendshipController.addFriend()`.
- Updated `FriendshipNotifier` to build friend request MQTT payloads internally.
- Added MQTT notifications for accepted and rejected friend requests.
- Added `FRIEND_REQUEST_ACCEPTED` and `FRIEND_REQUEST_REJECTED` notification types.
- Updated friendship accept/delete flows to send MQTT events only after successful state changes.
- Ensured rejected request MQTT events are sent only for pending requests rejected by the recipient.
- Updated MQTT notification contract tests for friendship request created, accepted, and rejected events.
- Removed the obsolete service-level friend request notification test.

**Notes**

- The implementation uses the shared MQTT envelope via `buildMqttNotification`, but the inner payload shape is friendship-specific and is built inside `FriendshipNotifier`. It does not introduce a shared cross-feature payload DTO.
- One old test file where MQTT was tested from service level is removed and now we have consistent controller-notifier-service implementation
- This is part 6 of issue 842


tickBit:
PR is ready now, only MQTT docs added afterwards.

CapoMK25 (approved):
Solid work! Approved.

PR #929, opened 2026-08-24
Added authorization to Box endpoints

### LINKED ISSUE #897: Require GroupAdmin authentication for reading Box endpoints
Currently, `GET /box` and `GET /box/{_id}` are accessible without authentication because they are marked with `@NoAuth()`. These endpoints were described as development-time endpoints, but upcoming classroom testing is no longer a development-only context.

These endpoints must be protected so that only an authenticated `GroupAdmin` can read Box data.

**Required Changes**
- Require valid GroupAdmin authentication for:
  - `GET /box`
  - `GET /box/{_id}`
- Remove unauthenticated access from both endpoints.
- A GroupAdmin must only be able to read their own Box.
- `GET /box` must not return Boxes belonging to other GroupAdmins.
- `GET /box/{_id}` must reject access if the requested Box does not belong to the authenticated GroupAdmin.
- Update API documentation metadata so these endpoints are shown as authenticated endpoints.

**Acceptance Criteria**
- Calling `GET /box` without an auth token returns `401`.
- Calling `GET /box/{_id}` without an auth token returns `401`.
- Calling either endpoint with a non-GroupAdmin testing-session token returns `403`.
- A GroupAdmin can read their own Box.
- A GroupAdmin cannot read another GroupAdmin’s Box.
- Tests cover unauthenticated access, non-GroupAdmin access, own-Box access, and cross-GroupAdmin access.

**Notes**
Existing GroupAdmin-protected Box endpoints use `@IsGroupAdmin()`, so the implementation should follow the same authentication and authorization pattern where appropriate.

Iir0R:
### Brief description
Added authorization checks to Get/box and Get/box/:_id

### Change list

- /box and /box/:_id now require valid Admin access token.
- /box now only returns Admin's Boxes
- Updated requestHelper to fix problem with embedded schemas
- Added boxRules

NOTE: boxRules test uses "any" due to RulesSetterAsync expecting subjectObj to be object (?).


src/authorization/rule/boxRules.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ModelName } from '../../common/enum/modelName.enum';
+import { ForbiddenException, NotFoundException } from '@nestjs/common';
+import { MongooseError } from 'mongoose';
+import { BoxDto } from '../../box/dto/box.dto';
+
+type Subjects = InferSubjects<any>;
+type Ability = MongoAbility<[AllowedAction | Action.manage, Subjects | 'all']>;
+export const boxRules: RulesSetterAsync<Ability, Subjects> = async (
+  user,
+  subject: any,
```
tickBit, line 19: Just wondering, if `any` type is just left out, would it work?

src/authorization/rule/boxRules.ts
```diff
@@ -0,0 +1,50 @@
...
+  action,
+  subjectObj: any,
```
tickBit, line 21: Same as above

tickBit (commented):
I can approve this, if this doesn't work without those `any` types; their usage seems to be justified in that case.

Iir0R:
subjectObj: AllowedSubject | undefined, expects a Dto, not sure if I can change that? Most rules that have subjectObj seem to pass an id though.

tickBit:
> subjectObj: AllowedSubject | undefined, expects a Dto, not sure if I can change that? Most rules that have subjectObj seem to pass an id though.

Ok, I'll give my approval 👍

tickBit (approved):
Approved 👍 

I'll update the branch, my today's changes have put the branch out of date from the base branch. It should be just a press-a-button thing.

PR #931, opened 2026-08-25
907 clarify matchmaking documentation

### LINKED ISSUE #907: Clarify matchmaking documentation
**Summary**

`gameType="matchmaking"` appears unnecessary in `POST /gameData/battle/start`.

The documented matchmaking flow is handled through the `/matchmaking` endpoints and MQTT events. It does not require clients to start a battle through `gameData/battle/start`.

**Current Behavior**

`POST /gameData/battle/start` still accepts `gameType="matchmaking"` and contains matchmaking-specific validation around `matchId`.

This creates confusion because it suggests that matchmaking clients may need to call this endpoint, even though the matchmaking flow already creates the match and provides the match ID through `MATCH_FOUND` / `MATCH_STARTED`.

**Expected Direction**

Remove the need for `gameType="matchmaking"` in `POST /gameData/battle/start`.

This endpoint should be documented as supporting only non-matchmaking battle starts.

**Suggested Outcome**

Document that `POST /gameData/battle/start` is not part of the normal matchmaking flow.

tickBit:
### Brief description

The title is a bit misleading, the point of this PR is to keep matchmaking flow handled via MQTT. Into `gameData.controller.ts` is added into comments, that matchmaking is handled through MQTT only.

Remove `matchmaking` as a supported `gameType` for `POST /gameData/battle/start`. Matchmaking battle flow now stays fully under the `/matchmaking` endpoints and MQTT events, while `casual` and `custom` battle starts remain unchanged.

Related issue: #907 Edit: In yesterday's meeting was ensured, that matchmaking is used via MQTT only on the backend.

### Change list

- Updated battle lifecycle tests to use `casual`/`custom` instead of `matchmaking` for `battle/start`.
- Added coverage to ensure `gameType="matchmaking"` does not create a battle through `battle/start`.
- Updated `StartBattleDto` validation tests to reject `matchmaking`.
- Restricted `StartBattleDto.gameType` runtime validation to `custom` and `casual`.
- Added service-level guard returning `WRONG_ENUM` for unsupported battle-start game types.
- Verified with `npm test -- gameData --runInBand` and `npm test -- matchmaking --runInBand`.

**Note**

I had to change in one older test file `let` to `const` to get any commits through.

### DIFF
--- a/src/__tests__/authorization/rule/boxRules.test.ts
+++ b/src/__tests__/authorization/rule/boxRules.test.ts
@@ -6,7 +6,7 @@ import { boxRules } from '../../../authorization/rule/boxRules';
 import { BoxDto } from '../../../box/dto/box.dto';
 
 describe('boxRules() test suite', () => {
-  let boxId = 'boxid';
+  const boxId = 'boxid';
 
   let requestHelperService: RequestHelperService;
   let getModelInstanceById: jest.Mock;
@@ -14,7 +14,7 @@ describe('boxRules() test suite', () => {
   const boxUserBuilder = BoxBuilderFactory.getBuilder('BoxUser');
   const boxAdminUser = boxUserBuilder.setGroupAdmin(true).build();
 
-  beforeEach( async () => {
+  beforeEach(async () => {
     getModelInstanceById = jest.fn();
     requestHelperService = {
       getModelInstanceById,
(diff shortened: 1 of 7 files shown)


CapoMK25 (approved):
Solid, approved! I was going to comment on the let --> const change, but the variable is almost certainly supposed to be a constant in the first place.

PR #935, opened 2026-08-26
908 bug calling post gamedatabattle causes runtime error

### LINKED ISSUE #908: Bug: Calling POST /gameData/battle causes runtime error
## What doesn't work?

Calling the legacy battle result endpoint after starting a battle can crash the API with a runtime error:

```text
TypeError: Cannot read properties of undefined (reading 'getTime')
```

The crash happens because `POST /gameData/battle/start` creates a battle document without `endedAt`, but the legacy `POST /gameData/battle` flow later finds that same document in `gameAlreadyExists()` and assumes `game.endedAt` exists:

```ts
if (game.endedAt.getTime() < currentTime.getTime() - 30 * 1000)
```

On dev, this may surface to clients as `502 Bad Gateway`.

Endpoint: `/gameData/battle`

Method: `POST`

Related endpoint: `/gameData/battle/start`

Related method: `POST`

## How can it be reproduced?

1. Start a battle with `POST /gameData/battle/start`, for example with `gameType: "casual"` and valid `team1` / `team2` player IDs.
2. Use the same teams to call the legacy result endpoint `POST /gameData/battle`.
3. Observe that the API can throw:

```text
TypeError: Cannot read properties of undefined (reading 'getTime')
```

## What should it do instead?

The API should not crash.

If `POST /gameData/battle` is still supported, it should ignore battle-start documents without `endedAt` when checking for duplicate completed games, or safely handle missing `endedAt`.

Possible fix direction:

```ts
if (!game.endedAt) return false;
```

or filter the query:

```ts
endedAt: { $exists: true }
```

If the intended flow is now `POST /gameData/battle/start` followed by `PUT /gameData/battle/result`, then `POST /gameData/battle` should return a clear `400 Bad Request` or documentation should explicitly mark it as legacy/deprecated.

There is a related flow gap: the new `PUT /gameData/battle/result` endpoint can move a battle to `COMPLETED`, but it does not currently return a `stealToken`. The old `POST /gameData/battle` legacy result flow is still the only path that generates the `stealToken` needed for `/item/steal`.

tickBit:
### Brief description

Handle case, where `endedAt` (`gameAlreadyExists` function in `gameData.service.ts`) field is not found and add `stealToken` support to the new `PUT /gameData/battle/result` flow.

### Postman testing

After adding code line `if (!game.endedAt) return false;`

```
{
    "data": {
        "Object": {
            "stealToken": "...",
            "soulHome_id": "6a06dc34058c7fd326e2b31b",
            "roomIds": [
                "6a06dc34058c7fd326e2b31d",
                "6a06dc34058c7fd326e2b31f",
				...
				]
```

Was: run time error.

`PUT /gameData/battle/result`: (1 vs. 1)

```
{
    "battleDocument": {
        "_id": "6a8ebce5f76e9f4af9642c32",
        "team1": [
            "69ef8b7332f46bd743f4c54e"
        ],
        "team2": [
            "69f485c5fbe1d439c999ac6a"
        ],
        "gameType": "casual",
        "status": "OPEN",
        "receivedResults": [
            {
                "playerId": "69ef8b7332f46bd743f4c54e",
                "winnerTeam": 1,
                "duration": 120,
                "_id": "6a8ebd00f76e9f4af9642c38",
                "receivedAt": "2026-08-26T10:16:32.469Z",
                "id": "6a8ebd00f76e9f4af9642c38"
            }
        ],
        "__v": 1,
        "id": "6a8ebce5f76e9f4af9642c32"
    },
    "stealToken": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwbGF5ZXJJZCI6IjY5ZWY4YjczMzJmNDZiZDc0M2Y0YzU0ZSIsInNvdWxIb21lSWQiOiI2YTA2ZGMzNDA1OGM3ZmQzMjZlMmIzMWIiLCJpYXQiOjE3ODc3MzkzOTIsImV4cCI6MTc4Nzc0MDI5Mn0.ABvXAEqPS9xvgivy-IFjYsooy0gOEr4I7vz_IgfmZuA",
    "soulHome_id": "6a06dc34058c7fd326e2b31b",
    "roomIds": [
        "6a06dc34058c7fd326e2b31d",
        "6a06dc34058c7fd326e2b31f",
```

### Change list

- Updated `handleBattleResult` to return steal data for valid winning result submissions.
- Added completed-battle response handling that includes `stealToken`, `soulHome_id`, and `roomIds`.
- Reused the legacy battle response generation logic for consistent steal payload behavior.
- Added fallback token generation when a losing player completes the battle, using a submitted winning player as the token owner.
- Updated JSDoc and inline comments for the new result/steal-token behavior.
- Updated `battleLifecycle.test.ts` to cover steal-token responses from `PUT /gameData/battle/result`.

### DIFF
--- a/src/__tests__/gameData/gameDataService/battleLifecycle.test.ts
+++ b/src/__tests__/gameData/gameDataService/battleLifecycle.test.ts
@@ -6,6 +6,7 @@ import { GameType } from '../../../gameData/enum/gameType.enum';
 import { BattleStatus } from '../../../gameData/enum/battleStatus.enum';
 import ServiceError from '../../../common/service/basicService/ServiceError';
 import { SEReason } from '../../../common/service/basicService/SEReason';
+import { JwtService } from '@nestjs/jwt';
 
 describe('GameDataService battle lifecycle test suite', () => {
   const gameDataModel = GameDataModule.getGameModel();
@@ -29,9 +30,24 @@ describe('GameDataService battle lifecycle test suite', () => {
     jest
       .spyOn(gameDataService.playerService, 'getPlayerById')
       .mockImplementation(async (playerId: string) => [
-        { _id: playerId } as any,
+        {
+          _id: playerId,
+          clan_id: new Types.ObjectId().toHexString(),
+        } as any,
         null,
       ]);
+    jest.spyOn(gameDataService.clanService, 'readOneById').mockResolvedValue([
+      {
+        SoulHome: {
+          _id: new Types.ObjectId().toHexString(),
+        },
+      } as any,
+      null,
+    ]);
+    jest
+      .spyOn(gameDataService.roomService, 'readAllSoulHomeRooms')
+      .mockResolvedValue([[] as any, null]);
+    jest.spyOn(JwtService.prototype, 'signAsync').mockResolvedValue('token');
   });
 
   it('Should register a battle using the provided matchId as document _id', async () => {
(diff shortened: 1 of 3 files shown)


CapoMK25 (approved):
Solid work, approved.

PR #937, opened 2026-08-26
920 bug put room allows unsafe room ownership updates and returns unclear 500 errors

### LINKED ISSUE #920: Bug: PUT /room allows unsafe room ownership updates and returns unclear 500 errors
## What doesn't work?

`PUT /room` has multiple room update issues:

1. A valid room update can return an opaque `500 UNEXPECTED` if MongoDB transactions are unavailable, for example when local MongoDB is not running as a properly initialized replica set.
2. Sending only `_id`, or sending only unknown update fields such as the old `floorType`, results in a `500` instead of a clear `400`.
3. The endpoint can allow moving an owned room to another clan's SoulHome by sending `soulHome_id` in the body.
4. The controller type suggests array/bulk updates are supported, but authorization does not handle array bodies correctly.

Endpoint: `/room`

Method: `PUT`

## How can it be reproduced?

1. Log in as a player who belongs to a clan and has access to a room.
2. Send:

```json
{
  "_id": "valid-own-room-id",
  "floor": "Stone"
}
```

3. If MongoDB is not running as a replica set, the response is `500 UNEXPECTED`.

Also reproducible with:

```json
{
  "_id": "valid-own-room-id",
  "floorType": "Stone"
}
```

or:

```json
{
  "_id": "valid-own-room-id"
}
```

These produce `500` because no valid update fields remain after validation/whitelisting.

Cross-clan issue:

```json
{
  "_id": "valid-own-room-id",
  "soulHome_id": "other-clan-soulhome-id",
  "floor": "Stone"
}
```

This can move the room under another clan's SoulHome.

**Important clarification:**

A player does not seem to be able to directly update another clan's existing room by sending that room's `_id`; that case returns `403 Forbidden`.

However, a player can update their own room and include `soulHome_id` in the body:
```
{
  "_id": "valid-own-room-id",
  "soulHome_id": "other-clan-soulhome-id",
  "floor": "Stone"
}
```
In this case, authorization checks only the room's current `soulHome_id` from the database, so the request is allowed because the room currently belongs to the player's clan. Later, the service writes all fields except `_id` with `$set`, including the body-provided `soulHome_id`. This can move the room under another clan's SoulHome.

Expected behavior: `soulHome_id` should not be updateable through `PUT /room`, and ownership/authorization should not rely only on the room's current owner if the update body can change ownership fields.

## What should it do instead?

- Return a clear `400 Bad Request` when no valid update fields are provided.
- Reject unknown or unsupported fields explicitly, or keep API field names aligned with the frontend.
- `PUT /room` should not blindly persist `soulHome_id` from the request body. If changing a room's SoulHome is supported at all, the target SoulHome must be explicitly authorized and must belong to the logged-in player's clan.
- Either properly support array/bulk updates in authorization or remove array support from the endpoint contract.

The controller currently accepts `UpdateRoomDto | UpdateRoomDto[]`, but authorization assumes a single object and reads `subjectObj._id`.

Example:
```
[
  {
    "_id": "valid-own-room-id",
    "floor": "Stone"
  }
]
```

This fails authorization because the body is an array, so `subjectObj._id` is `undefined`. Either array/bulk updates should be handled explicitly in authorization, or the endpoint contract should only allow a single `UpdateRoomDto`.

- Return a clearer setup/configuration error if MongoDB transactions are unavailable.

The frontend may still be sending `floorType`, while the backend DTO currently expects `floor`. Unless this rename has been agreed with the frontend, `UpdateRoomDto` should probably use `floorType` again, or support backwards compatibility during the transition.

Note about local MongoDB: this endpoint uses MongoDB transactions. Transactions require MongoDB to run as a replica set or through mongos. If local Docker MongoDB is not properly initialized as replica set `rs0`, even a valid update body such as `{ "_id": "...", "floor": "Stone" }` can return an opaque `500 UNEXPECTED` with an underlying MongoDB error like `Transaction numbers are only allowed on a replica set member or mongos`.

This may be an environment/setup issue, but the API response is currently not very actionable.

constf03:
### Brief description

PUT /room accepted a body type as "UpdateRoomDto | UpdateRoomDto[]" . According to Claude the global ValidationPipe (found in main.ts) was silently skipped for this endpoint so no validation or field whitelisting ever ran which should explain some of the issue's reported bugs:

- unknown or empty update bodies reaching Mongo unfiltered and returning a unclear 500 error 

- soulHome_id being "smuggled" through a room update to move it into another clan SoulHome

- also authorization rules breaking on array bodies because it only reads the "subjectObj._id " of a single object and not a bulk/array (this one not necessarily related to the ValidationPipe).

Removing bulk/array support from PUT /room was the safest choice because 1. the put /room endpoint doesn't seem to properly support array updates to begin with (neither the tests indicate that) and 2. re-writing the authorization rule for array support seems more complex and could cause ohter problems...

Also in this PR I clarified the unclear 500 returned when MongoDB is not running as a replica set or mongos. And renamed "floor" to "floorType" everywhere across the backend.

### Change list

- room.controller.ts: changed PUT /room body type from UpdateRoomDto | UpdateRoomDto[] to a single UpdateRoomDto restoring proper ValidationPipe handling and removing the array/bulk contract

- room.controller.ts: added a built in NestJS ValidationPipe (forbidNonWhitelisted: true) on the update route so unknown fields (including soulHome_id) are rejected with a clear 400 error status

- room.service.ts (updateSoulHomeRooms): added 400 SE REQUIRED error when an update has no actual fields to change instead of failing with a 500

- room.service.ts (updateSoulHomeRooms): simplified the method to operate on a single room instead of an array and stopped writing the furniture key into the Room document's $set

- room.service.ts (deactivateRoom) and join.service.ts: use floorType matching the current schema field name.

- BasicService.ts (convertMongooseToServiceErrors ): map MongoDB's " not a replica set / mongos" transaction error to a ServiceError instead of a generic UNEXPECTED one

- ValidationExceptionFilter.ts: added new apiValidationExceptionFactory as a reusable exceptionFactory for NestJS ValidationPipe that is now used by the PUT / room controller (should reject fields that are not in the UpdateRoomDto like soulhome_id)

- added and updated tests in updateSoulHomeRooms.test.ts covering the new REQUIRED error, the null/undefined guard and furniture handling

- renamed room field `floor` back to `floorType` in all the files where it's present.

### Extra

While working on this issue a data loss bug was found in `updateSoulHomeRooms` . The method compares the furniture array in the request body against the items currently in the room and unassigns to stock any item not present in that list. When furniture was omitted from the request body it was treated the same as an empty array which means a request body for /PUT Room like

`{ "_id": ..., "roomColour": "blue" }`

would silently remove every item out of the room. This was probably a overlooked in the previous PUT /room furniture-related issue I think...

Now not including furniture in the request body should leaves the room's items untouched. This is Covered by an updated test that places an item in the room before an update without furniture and asserts the item is still there after.


tickBit:
I read the PR text: amazing work 👍 ! Regarding the furniture, really nice addition 👍 

The branch seems to be out-of-date with the base branch. Please update the branch and I will try to give a review today.

tickBit:
If the CI pipeline is stuck, an "empty" commit might help. I used that trick once this week and it worked.

tickBit:
Seems to be still stuck. Does @CapoMK25 have any idea how we could get the CI to run the tests?

CapoMK25:
A couple. Closing the PR completely and reopening is the fastest one, but also the worst one. 

The second is using a couple of flags on top of pushing an empty commit for GH Actions. I'll do this next, and go from there.

CapoMK25:
I'll take a look at Actions on the repo itself as well, looks like the runners are genuinely stuck.

CapoMK25:
https://www.githubstatus.com/

Should'e checked this one first, but I wanted to null out the usual suspects. So Actions seems to be having issues currently, 3PM UTC matches to 6PM Finnish time. @tickBit @constf03

tickBit (approved):
Good work 👍 ! Approved.

Issue #936, opened 2026-08-26
Feature request: Support configurable daily tasks with optional rewards


tickBit:
## What do you want to see in the API?

Add support for server-controlled daily task availability using a minimal daily task config object.

The API should continue returning the active daily tasks for the logged-in player's clan from `GET /dailyTasks`. The response should include the daily tasks that have been generated for that clan. It should not simply return every value from the task enums.

Each generated task should be based on a daily task config where `name` is the only required field:

- `name: ServerTaskName | UITaskName`
- `rewardPoints?: number`
- `coins?: number`

Example minimal config:
```
{
  "name": "play_battle"
}
```

Example config with overrides:
```
{
  "name": "play_battle",
  "rewardPoints": 150,
  "coins": 75
}
```
The following new task names should be added to the appropriate daily task enums if they do not already exist:

form_an_inner_connection
play_with_emotions
letting_go_of_the_old
your_voice
inner_discussion
recycling_experiences
inner_voice
choose_your_truth
hold_the_character
where_are_you
follow_the_voice
build_your_world
read_the_signs
find_the_roots
look_inside
set_boundaries
use_of_power
common_factor
spiritual_currency
reward_traps
moral_dilemmas
eco_friendly
hunting_for_values
ethics_metrics

The following provided task names seem to already have existing or close equivalents and should be reviewed before adding duplicates:

banish_the_earworm -> change_song_soulhome
go_to_battle -> play_battle
stronger_soldier -> change_character_be_strong
find_the_error -> find_bug
play_melody -> make_music_with_buttons

## How do you think this should work?

The client/game already has its own implementation and default data for each daily task. The server should not need to send the full task definition. Instead, the server decides which tasks should be active by generating and storing task configs for each clan.
The minimum required task config is:
```
{
  "name": "play_battle"
}
```

If the server includes optional fields, those values should override the defaults. For example, if the default task reward is 20 points but the server sends `rewardPoints: 150`, the generated task should use 150 points.
`coins` should also be optional because different daily tasks may reward different amounts of game currency.

Suggested behavior:
```
const generatedTask = {
  ...defaultTaskValues,
  ...serverTaskConfigOverrides,
};
```

`GET /dailyTasks` should then return the generated tasks from the database for the logged-in player's clan, as it does currently. This endpoint should include both UI-managed tasks and server-managed tasks if both have been generated for the clan.
The current behavior is conceptually correct:
- UI tasks are generated from the predefined `uiDailyTasks` record.
[...]

PR #944, opened 2026-08-31
936 new daily tasks with startup refresh check

tickBit:
### Brief description

Adds temporary startup maintenance refreshes for cleaning up old persisted data without requiring direct MongoDB access or exposing server credentials. The changes are intentionally isolated in separate files so they can be removed easily once the physical server data has been refreshed.

Related issue #936 . The issue description is a bit misleading. The point of this PR is to have new daily tasks gradually to the project.

### Change list

- Added a temporary daily task startup refresh that checks persisted `DailyTask` documents for task types listed in `oldTaskNames.enum.ts`.
- Added MongoDB-based locking via `MaintenanceLock` so multiple backend instances cannot run the daily task refresh at the same time.
- Regenerates daily tasks only when old task names are found.
- Uses the current UI task definitions and temporary server task generation for refreshed clan tasks.
- Updated battle daily task logic so `PLAY_BATTLE` and `WIN_BATTLE` are logically replaced by one task: `GO_TO_BATTLE`.
- Added separate, easily removable tests for the daily task startup refresh.
- Added a temporary Room startup refresh for converting old `floor` fields to `floorType`.
- Added separate, easily removable tests for the Room `floor` -> `floorType` refresh.
- Uses Nest `Logger` instead of `console.*`.

### Extra

These maintenance services are temporary migration helpers. They are kept separate from normal request handling and ordinary domain logic to make future removal straightforward: remove the startup refresh service files, their module provider registrations, and their matching test directories.

Over time, **some of the currently active server tasks are expected to become legacy tasks** and move into `oldTaskNames.enum.ts`. The daily task refresh is designed around that enum so the cleanup trigger remains explicit and easy to audit.

The Room migration is included as an extra cleanup step for physical server data, converting persisted `floor` values into the current `floorType` field while preserving existing `floorType` values when already present.


tickBit:
When all the new daily tasks are in place, the files to remove are:

- `__tests__/DailyTasks/DailyTasksStartupRefreshService/startupRefresh.test.ts`
- `dailyTasks/enum/oldTaskNames.enum.ts`

- `dailyTasksStartupRefresh.service.ts`
and `dailyTasks.module.ts` doesn't need to import the above anymore

Similar for Room refresh.

CapoMK25 (approved):
Solid, approved.

PR #946, opened 2026-09-01
Fixed Box data cleanup

### LINKED ISSUE #919: Feature request: Clean up after PUT /box/reset

After a GroupAdmin resets a box session and signs in again, starting a new session with the new access token should work without hitting conflicts from data created by the previous session.

Currently, the GroupAdmin re-login itself works correctly after reset, but calling `POST /box/start` again can return `409` because the API tries to create a clan with a name that already exists from the previous session.

`PUT /box/reset` should fully clean up all session-specific data created during the previous testing session, including the clans created by `POST /box/start` and their related data.

Expected flow:

1. `POST /box/createAdmin` returns `204`
2. `POST /box` returns `201`
3. `POST /box/start` returns `204`
4. `PUT /box/reset` returns `204`
5. `POST /box/admin/signIn` returns `201` with a new `accessToken`
6. `POST /box/start` with the new token should return `204` and start a fresh session

The reset should leave the box in a clean `PREPARING` state where a new session can be started without duplicate clan name conflicts.

## Any additional info?

Observed after reset and re-login:

`POST /box/start` returned `409` with:

```text
Field "name" with value "Urheat-Ilvekset" already exists.
```

This suggests that the previous session’s created clan was not removed, or it was not associated with the box in a way that the reset cleanup can find and delete. The GroupAdmin re-login appears to work as expected; the issue seems to be incomplete cleanup of session-created data.

Iir0R:
### Brief description
Fixed BoxService reset not removing all data.
NOTE: Works only when using TESTING_SESSION

### Change list

- Made all methods use same session to fix some data not receiving box_id
- Fixed getting all collections in clearSession
- Added 'Game' to BoxCommonModule
- Updated UpdateClanDto to contain optional box_id to get rid of 'any'
- Added 'box_id' to box creation for consistency


src/box/sessionStarter/sessionStarter.service.ts
tickBit: This seems to be from old implementation. I think here could be `return { ...dailyTask, clan_id: clan1_id as string | ObjectId };` or something like that.

tickBit: Same as above

tickBit:
Since the codecov/project seems to be stuck, you could try replacing those old `any` types with `string | ObjectId`, one commit could help to resolve the jam of the system. But if changing `any` => `string | ObjectId` causes unexpected extra problems, one "empty" commit that can be "anything" might help to get the system going.

Overall, seems good to me.

Edit: Sorry for earlier typos.

tickBit (approved):
Nice! Good gob 👍 ! Approved.

Since I seem to be a bit tired, I'll let @CapoMK25 take a look, too.

CapoMK25 (approved):
Approved!

PR #953, opened 2026-09-02
942 Fixed Box admin sign in, added Testing session guard

### LINKED ISSUE #942: Bug: GroupAdmin re-login does not work in TESTING_SESSION environment
## What doesn't work?

GroupAdmin re-login from `/box/admin/signIn` works, if in environment settings PRODUCTION is set. If the environment is TESTING_SESSION, Postman gives with valid credentials:
```
{
    "errors": [
        {
            "response": "NOT_AUTHENTICATED",
            "status": 401,
            "message": "All endpoints need to be provided an auth token in testing sessions.",
            "name": "APIError",
            "reason": "NOT_AUTHENTICATED",
            "field": null,
            "value": null,
            "additional": null,
            "statusCode": 401
        }
    ],
    "statusCode": 401
}
```
In TESTING_SESSION mode GroupAdmin should be able to re-login given then a box with token.

Endpoint: 

`/box/admin/signIn`

Method: 

`POST`

## How can it be reproduced?

1. Set the server to run in TESTING_SESSION
2. Start the server
3. Clear Authorization field in Postman
4. Try logging in with correct GroupAdmin credentials using Postman
5. The endpoint will return status code 401

## What should it do instead?

The endpoint should work just like it does in PRODUCTION mode currently: return data in form
```
{
    "data": {
        "Box": {
            "accessToken": ..."
        }
    },
    "metaData": {
        "dataKey": "Box",
        "modelName": "Box",
        "dataType": "Object",
        "dataCount": 1
    }
}
```

With current information I have, it seems, that GroupAdmin re-login is meant for TESTING_SESSIONs only. That is, why I think, it is safest to remove the ability for GroupAdmin to re-login, if environment is not TESTING_SESSION.

Iir0R:
### Brief description
Fixed Box Testing session Admin sign in. Added Guard to prevent use of endpoints outside Testing Session, due to many of them requiring box_id, which is only added in testing.

### Change list

- Added NoBoxIdFilter to box/admin/SignIn
- Created BoxTestingSessionGuard and added to Box related Controllers
- Tests for Guard


CapoMK25 (approved):
Solid in my view all around, approved.

tickBit (approved):
Good work 👍 ! Approved.

PR #961, opened 2026-09-05
Modified auth interceptor and added a test on top

### LINKED ISSUE #955: Bug: Multiple room updates all at once doesn't work
## What doesn't work?

`PUT /room` is expected to support updating either a single room object or an array of room update objects.

The service and room authorization rule both appear to support arrays, but the request fails when sent through the actual HTTP endpoint, for example from Postman. The response is a 403 error:

`The logged-in user has no permission to execute update_request action`

The likely bug is in `AuthorizationInterceptor`. For update requests it does:

`plainToInstance(subject, request.body)`

When `request.body` is a top-level array, this returns an array of `UpdateRoomDto` instances. CASL then checks permission against the array itself, whose subject type is `Array`, not `UpdateRoomDto`. Because the ability rule grants `update_request` for `UpdateRoomDto`, the check fails even though the user has permission to edit the rooms.

There is likely a second issue in the same update block: after the permission check, the interceptor does:

`request.body = pick(dataClass, allowedFields)`

For an array payload, this does not preserve the array of update objects correctly.

The existing tests do not catch this because they appear to test `RoomService.updateSoulHomeRooms()` directly and/or `roomRules()` directly, but not the full HTTP request path through `AuthorizationInterceptor`.

Endpoint: `/room`

Method: `PUT`

## How can it be reproduced?

1. Log in as a user who belongs to the same clan as the rooms being updated and has room edit permissions.

2. Send a `PUT /room` request with a top-level JSON array body, for example:

```json
[
  {
    "_id": "ROOM_ID_1",
    "floorType": "wood"
  },
  {
    "_id": "ROOM_ID_2",
    "wallpaper": "painted"
  }
]
```

3. Observe that the request fails with `403 Forbidden`.

4. The response message is:

```text
The logged-in user has no permission to execute update_request action
```

Sending a single room update object does work, which indicates the failure is specific to the top-level array payload passing through the authorization interceptor.

CapoMK25:
### Brief description

Issue #955 

I added a test to check that the behavior doesn't repeat again on top of doing the changes to the file. The functionality in the authorization interceptor file is now split into two use cases: the single object logic that was present originally, and the array logic that's added as a new addition on top of that. 

### Change list

- Updated `AuthorizationInterceptor` (`src/authorization/authorization.interceptor.ts`) to detect top-level array payloads `(Array.isArray(request.body))` during `UPDATE` requests.

- Refactored permission evaluation to iterate over array items individually with `plainToInstance()`, ensuring CASL checks permissions against individual DTO instances rather than the outer Array container type.

- Applied `lodash.pick()` per item in array bodies to prevent field stripping and maintain correct array structure.

- Preserved original single-object handling logic for standard non-array update payloads.

- Added a new unit test suite (`src/__tests__/authorization/authorizationInterceptor.array.test.ts`) to provide testing coverage for top-level array updates passing through the authorization interceptor.

### DIFF
--- a/src/__tests__/authorization/authorizationInterceptor.array.test.ts
+++ b/src/__tests__/authorization/authorizationInterceptor.array.test.ts
@@ -1,0 +1,77 @@
+import { AuthorizationInterceptor } from '../../authorization/authorization.interceptor';
+import { ExecutionContext } from '@nestjs/common';
+import { of } from 'rxjs';
+import { User } from '../../auth/user';
+import { createMongoAbility, AbilityBuilder } from '@casl/ability';
+
+describe('AuthorizationInterceptor - Array Payload Handling', () => {
+  let interceptor: AuthorizationInterceptor;
+  let mockCaslFactory: any;
+  let mockReflector: any;
+
+  // Define the DTO class for the test
+  class TestDto {
+    _id!: string;
+    name!: string;
+  }
+
+  beforeEach(() => {
+    const { can, build } = new AbilityBuilder(createMongoAbility);
+    
+    // Grant permissions here for the test user
+    can('update_request', TestDto);
+    const realAbility = build();
+
+    mockCaslFactory = {
+      createForUser: jest.fn().mockResolvedValue(realAbility),
+    };
+
+    mockReflector = {
+      get: jest.fn().mockReturnValue({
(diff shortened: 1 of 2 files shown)


tickBit (approved):
Looks good to me. Approved.

PR #964, opened 2026-09-07
current linter & Prettier changes only

tickBit:
### Brief description

Current linter and Prettier changes only


CapoMK25 (approved):
Auto-approved!

PR #970, opened 2026-09-09
Bug fix: endedAt doesn't appear after 10 minutes

EemeliJ:
### Brief description

Fix expired failed clan-role votings remaining unfinished.
Previously, set_clan_role votings returned early when they failed to reach the required majority, so finalizeVoting() was never called and endedAt remained unset. Now all expired role votings are finalized, while player roles are updated only when the vote passes.

### Change list

- Updated ClanRoleService.checkVotingOnExpire(): 
Moved player role updates into the succesful-vote branch.
Calls finalizeVoting() for both passed and failed votings.

- Updated clan-role expiry tests:
Verify finalization is called after both succesful and failed votes.

- (extra) Swagger sign-in documentation
Added and explicit request body schema for POST/auth/signIn (For testing through Swagger).


tickBit (approved):
Excellent work 👍 ! Approved.

I tested locally with this code, that if the voting doesn't pass with majority, `endedAt` appears and if it does pass with majority, `endedAt` appears, too and the voting result comes into place. Everything works as excepted.

PR #981, opened 2026-09-12
Add INNER_VOICE clan task and balanced daily-task generation

### LINKED ISSUE #967: Feature request: Implement INNER_VOICE daily task

Add a server-tracked daily task: `INNER_VOICE`.

The task should be completed when a player successfully updates and saves the clan phrase/motto.

The player-facing instruction for the task is:

```json
{
  "fi": "Avaa klaanin asetukset. Muokkaa klaanin mottoa ja tallenna muutos. Mieti, mitä haluatte viestiä toisillenne ja muille."
}
```

The task should use the existing clan `phrase` field.

The task should be configured with:

```ts
{
  type: ServerTaskName.INNER_VOICE,
  amount: 1,
  amountLeft: 1
}
```

The frontend opens the clan settings UI and lets the player edit the clan motto/phrase.

When the player saves a changed `phrase` value successfully through the backend, the server should progress or complete the `INNER_VOICE` daily task.

Expected flow:

1. Player opens clan settings.
2. Player edits the clan motto/phrase.
3. Frontend sends the updated `phrase` value to the backend.
4. Backend validates and saves the updated `phrase`.
5. Backend progresses/completes the `INNER_VOICE` daily task.
6. Frontend receives the updated clan data normally through the API and/or MQTT.

## About MQTT

When the clan phrase is successfully updated, the backend should publish a clan update MQTT notification so clan members can update their local clan data without polling.

Suggested payload:

```ts
{
  clan_id: string;
  phrase: string;
}
```

Example:

```json
{
  "clan_id": "67fe4e2d8a54d4cc39266a43",
  "phrase": "Together we rise"
}
```

The MQTT event should be sent only after the phrase update has been successfully saved.

## Any additional info?

This should be treated as a clan-level daily task.

Changing the clan phrase/motto affects the whole clan, so completing `INNER_VOICE` should reward/progress the clan task rather than only an individual player task.

`INNER_VOICE = 'inner_voice'` already exists in `ServerTaskName`.

The task must be included in `taskGenerator.service.ts`.

`dailyTasksStartupRefresh.service.ts` refreshes the daily tasks into MongoDB, so this task must also be included in the startup refresh generation path.

No separate enum for clan phrase/motto appears to exist currently. The existing `phrase` field is a `string`, so this task should track changes to that field.

The clan phrase/motto is not free text in the UI.

The frontend has a hardcoded list of allowed motto strings and their localizations. The player selects one of these predefined mottos, and the frontend sends the selected motto as a plain string to the backend through the existing `phrase` field.

The string may be Finnish or English, for example `"Victory or nothing!"`, but the frontend can still map it back to the localized Finnish and English versions because all allowed mottos are known and hardcoded on the frontend side.

The backend does not need a separate enum for clan mottos for this task. The backend should treat `phrase` as the existing string field, save the received value, and complete/progress `INNER_VOICE` when the phrase has been successfully changed.

### LINKED ISSUE #974: Feature request: Daily server task generation should guarantee balanced coverage of all active server daily task types

Currently, server daily tasks are selected independently with random choice, so with bad luck some task types may appear only once or not at all during a daily reset. We need a scalable generation system where every active daily task type appears more than once, while still keeping the final task list randomized.

Replace pure random selection with a balanced “task bag” system.

Suggested approach:

- Keep all active server daily task definitions in one registry/config.
- Define a minimum occurrence count, for example `MIN_OCCURRENCES_PER_TASK_TYPE = 2`.
- Build an initial task bag by adding each active task type to the bag at least that many times.
- Shuffle the bag, for example with a Fisher-Yates shuffle.
- If more task slots are needed, fill the remaining slots randomly from the active task pool, optionally using weights.
- Generate final task values such as `amount`, `title`, `coins`, and `timeLimitMinutes` from the selected task definitions.

The generator should validate that:

```ts
SERVER_TASKS_PER_CLAN >= activeTaskTypes.length * MIN_OCCURRENCES_PER_TASK_TYPE
```

If this is not true, task generation should fail clearly, because the guarantee cannot be fulfilled.

The same generator should be used by both scheduled daily resets and startup refresh logic, so the behavior does not drift between code paths.

## Any additional info?

The current implementation has duplicated random generation logic in `taskGenerator.service.ts` and `dailyTasksStartupRefresh.service.ts`.

This change should make adding or removing daily task types easier: future task changes should happen mainly by updating the active task definition registry, not by editing random selection logic or multiple switch/case blocks.

Completed-task replacement should also be considered. If completed tasks are immediately replaced with a purely random task, the initially balanced distribution may drift during the day. A follow-up or part of this issue should decide whether replacements should prefer currently underrepresented task types.

**Extra**

Example of Fisher-Yates shuffle:
```
function shuffle<T>(items: T[]): T[] {
  const shuffled = [...items];

  for (let i = shuffled.length - 1; i > 0; i--) {
    const j = Math.floor(Math.random() * (i + 1));
    [shuffled[i], shuffled[j]] = [shuffled[j], shuffled[i]];
  }

  return shuffled;
}
```

EemeliJ:
### Brief description
(closes #967 & #974 )

Adds the clan-level INNER_VOICE daily task and replaces random initial server-task generation with a balanced shuffled task bag. Updating a clan’s phrase to a new value completes INNER_VOICE, rewards the clan, and publishes a clan MQTT update after the database transaction commits.

### Change list
- Add INNER_VOICE to the active server daily-task registry.
- Generate 11 server tasks per clan with every active type appearing at least twice.
- Use the shared balanced generator for new clans and startup refresh.
- Validate that the configured task-slot count can satisfy the minimum-occurrence rule.
- Progress INNER_VOICE when an authorized player saves a genuinely changed clan phrase.
- Make phrase save, task completion, and clan rewards transactional.
- Prevent individual player rewards for clan-level INNER_VOICE completion.
- Publish CLAN_UPDATED MQTT notifications with { clan_id, phrase } after commit.
- Add tests for balanced generation, phrase-triggered completion, no-op phrase updates, transaction rollback, clan rewards, and MQTT payloads.


tickBit (approved):
Great work 👍 ! Approved.

I looked carefully and to get context looked also other files and noticed a bug outside this PR. In `dailyTask.queue.ts` at line 56 there is field `playerId`, it should be `player_id` to clear the ID:
```
$unset: {
  player_id: '',     // Correct, but at the moment there is playerId
  startedAt: '',
}
```
Just wondering, that could you also change that field name in `dailyTask.queue.ts` in this PR? It's one field in one line only.

The branch was out of date, I updated it from the UI, so refreshing the branch locally is necessary. I'll wait until Monday morning before merging. That other fix isn't required. It can be fixed in other PR, too.

EemeliJ:
> Great work 👍 ! Approved.
> 
> I looked carefully and to get context looked also other files and noticed a bug outside this PR. In `dailyTask.queue.ts` at line 56 there is field `playerId`, it should be `player_id` to clear the ID:
> 
> ```
> $unset: {
>   player_id: '',     // Correct, but at the moment there is playerId
>   startedAt: '',
> }
> ```
> 
> Just wondering, that could you also change that field name in `dailyTask.queue.ts` in this PR? It's one field in one line only.
> 
> The branch was out of date, I updated it from the UI, so refreshing the branch locally is necessary. I'll wait until Monday morning before merging. That other fix isn't required. It can be fixed in other PR, too.

I fixed this naming mismatch, but i also spotted another similar one at dailyTasks.schema.ts at line 54: `DailyTaskSchema.index({ playerId: 1 }, { unique: true, sparse: true }); ` Should i apply the same fix to this one too? @tickBit

tickBit:
> > Great work 👍 ! Approved.
> > I looked carefully and to get context looked also other files and noticed a bug outside this PR. In `dailyTask.queue.ts` at line 56 there is field `playerId`, it should be `player_id` to clear the ID:
> > ```
> > $unset: {
> >   player_id: '',     // Correct, but at the moment there is playerId
> >   startedAt: '',
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Just wondering, that could you also change that field name in `dailyTask.queue.ts` in this PR? It's one field in one line only.
> > The branch was out of date, I updated it from the UI, so refreshing the branch locally is necessary. I'll wait until Monday morning before merging. That other fix isn't required. It can be fixed in other PR, too.
> 
> I fixed this naming mismatch, but i also spotted another similar one at dailyTasks.schema.ts at line 54: `DailyTaskSchema.index({ playerId: 1 }, { unique: true, sparse: true }); ` Should i apply the same fix to this one too? @tickBit

Good observation! I missed that. Yes, I think, that it should be `DailyTaskSchema.index({ player_id: 1 }, { unique: true, sparse: true });`. I looked, that GET /dailyTasks shows the field as `player_id`.

tickBit:
Nice! Now all all tests are green 👍 .

I'll merge this now.

EemeliJ:
Failing daily-task CI tests?

Root cause: clan and unreserved tasks use player_id: null. The previous unique sparse index still indexed null, so MongoDB allowed only one such task and rejected subsequent task creation as a duplicate key.  
The index is now a partial unique index: uniqueness applies only when a task is reserved by an actual player (player_id is an ObjectId). Clan/unreserved tasks can now coexist normally.

NOTE: A one-time production database maintenance step is required after deployment. The database currently has an old daily-task rule that treats every clan/unreserved task as the same because they have no player ID. A database admin must replace that rule with the new one, which allows multiple clan/unreserved tasks while still allowing each player to reserve only one task at a time.

PR #984, opened 2026-09-15
fix: handle chat message before websocket initialization

### LINKED ISSUE #965: Bug: Chat message before WebSocket user initialization causes internal server error
## What doesn't work?

Sending a chat message immediately after the WebSocket connection opens can fail with an internal server error.

The WebSocket is already open from the client’s perspective, but the backend may not have finished initializing `client.user` yet. If the client sends a `clanMessage` before `client.user` is set, the backend throws:

```text
TypeError: Cannot read properties of undefined (reading 'clanId')
```

Endpoint: `ws://localhost:8080/ws/chat`

Method: WebSocket event `clanMessage`

## How can it be reproduced?

1. Start the backend locally.
2. Sign in and get a valid JWT token.
3. Open a WebSocket connection to `ws://localhost:8080/ws/chat` using the token.
4. Send a `clanMessage` immediately in the client’s `open` handler, without waiting.
5. The server may respond with `Internal server error` and log a `TypeError` because `client.user` is still undefined.

## How should it be fixed?

The backend should not handle chat events before the WebSocket user context has been initialized.

Suggested changes:

1. Add an explicit guard/check at the beginning of chat event handlers such as `handleClanMessage`, `handleGlobalMessage`, and reaction handlers:

```ts
if (!client.user) {
  throw new WsException('WebSocket user is not initialized yet');
}
```

2. Return a controlled WebSocket error instead of allowing a `TypeError`.

3. Optionally send a `ready` event to the client at the end of `handleConnection()`, after the player has been loaded and `client.user` has been assigned. The client can then wait for this event before sending chat messages.

Example:

```json
{
  "event": "ready",
  "data": true
}
```

This removes the need for arbitrary client-side delays and makes the chat protocol deterministic.

**Extra:**

The following flow can be used to send a chat message in Powershell:

1. `$body = @{  username = "--username--" ; password = "--password--" } | ConvertTo-Json`
`

2. `$res = Invoke-RestMethod -Method Post -Uri "http://localhost:8080/auth/signIn" -ContentType "application/json" -Body $body`

3. `send-chat.js`:
```
const WebSocket = require("ws");

const token = process.argv[2];

if (!token) {
  console.error("Token puuttuu. Aja: node .\\send-chat.js SINUN_TOKEN");
  process.exit(1);
}

const sock = new WebSocket("ws://localhost:8080/ws/chat", {
  headers: {
    Authorization: "Bearer " + token,
  },
});

sock.on("open", () => {
  console.log("connected");

  setTimeout(() => {
    sock.send(JSON.stringify({
      event: "clanMessage",
      data: {
        content: "Testiviesti lokaaliymparistosta",
      },
    }));
  }, 1000);

  setTimeout(() => {
    sock.close();
  }, 3000);
});

sock.on("message", (message) => {
  console.log(message.toString());
});

sock.on("error", (error) => {
  console.error(error);
});

sock.on("close", () => {
  console.log("closed");
});
```

4. `node .\send-chat.js $env:TOKEN`

Please don't include `send-chat.js` into commits to repo.

EemeliJ:
### Brief description
Prevent chat events from causing an internal server error when sent before WebSocket user initialization completes.

### Change list
- Send a ready WebSocket event after the user context and chat rooms are initialized.
- Guard all chat message and reaction handlers against an uninitialized WebSocket user.
- Return a controlled WsException error instead of allowing a TypeError.
- Add gateway tests for readiness and uninitialized chat events.

closes #965


tickBit (approved):
Seems good solution to me 👍 . Approved.

PR #987, opened 2026-09-16
945 Added refresh token

### LINKED ISSUE #945: Feature request: Refresh token support for the normal auth flow

Add backward-compatible refresh token support for the normal auth flow.

This should affect:

```http
POST /auth/signIn
POST /auth/refresh
```

`POST /auth/signIn` should keep returning the current fields, including:

```json
{
  "accessToken": "...",
  "tokenExpires": 1234567890
}
```

It should additionally return:

```json
{
  "refreshToken": "...",
  "refreshTokenExpires": 1234567890
}
```

Add a new endpoint:

```http
POST /auth/refresh
```

Request body:

```json
{
  "refreshToken": "..."
}
```

Response:

```json
{
  "accessToken": "...",
  "tokenExpires": 1234567890,
  "refreshToken": "...",
  "refreshTokenExpires": 1234567890
}
```

When a user signs in through `POST /auth/signIn`, the API should create both an access token and a refresh token.

The access token should work exactly as it does now. The refresh token should be a longer-lived JWT that can be used to get a new access token without asking the user to sign in again.

Suggested refresh token payload:

```ts
{
  profile_id,
  player_id,
  type: 'refresh',
  tokenVersion
}
```

`tokenVersion` can use the existing field in the `Profile` schema. This allows old refresh tokens to be invalidated when needed, for example after password reset.

`POST /auth/refresh` should:

1. verify the refresh token signature and expiration
2. check that `type === 'refresh'`
3. find the profile by `profile_id`
4. check that `profile.tokenVersion ?? 0` matches the token payload
5. check that the player still exists
6. return a new access token and a new refresh token

Invalid, expired, or wrong-type tokens should return `401 Unauthorized`.

The old frontend should not break. It can ignore the new `refreshToken` and `refreshTokenExpires` fields until refresh token support is implemented on the frontend.

## Any additional info?

This issue should only cover the normal auth flow.

Do not change the Box/testing-session auth flow in this issue:

```txt
src/auth/box/BoxAuthService.ts
src/box/**
```

Files likely involved:

```txt
src/auth/auth.controller.ts
src/auth/auth.service.ts
src/auth/dto/signInResponse.dto.ts
src/auth/dto/refreshToken.dto.ts (NEW)
src/profile/profile.schema.ts
src/profile/profile.service.ts
```

`POST /profile` should not be changed to return tokens, because it currently creates a profile but does not sign the user in.

`POST /profile/guest` may receive refresh token support automatically because it calls `authService.signIn(...)` after creating the guest profile.

Suggested refresh token lifetime:

```txt
JWT_REFRESH_EXPIRES_IN=30d
```

Using environment variable is good practice, but while adding refresh token expiration, also make access token expiration use `envVars.JWT_REFRESH_EXPIRES_IN ?? '30d'` for consistency. Currently the expiration time of the token is hard coded. So, in the code the system could try to read the expiration times from .env, but if the field(s) don't exist, a hard coded value is used.

Add tests for:

```txt
POST /auth/signIn returns the old token fields
POST /auth/signIn also returns refreshToken and refreshTokenExpires
POST /auth/refresh returns new tokens with a valid refresh token
POST /auth/refresh rejects an access token
POST /auth/refresh rejects an invalid or expired token
POST /auth/refresh rejects a token if tokenVersion no longer matches
```

Password reset already increments `tokenVersion`, so old refresh tokens should stop working after password reset.

Iir0R:
### Brief description
Added refresh token for normal auth flow

### Change list

- New endpoint: auth/refresh/
- New service: refresh, validates refresh token data, increments tokenVersion to invalidate old token
- Changed createTestingSessionTokens methods name createTokens
- Updated signIn to use createTokens
- New tests + updated old


tickBit (approved):
Looks very good to me 👍 ! Approved.

